PR #174, opened 2019-08-08
update handler for DKG state breakup

evanmcc:
miner support for https://github.com/helium/erlang-dkg/pull/27


Vagabond:
Looks like there's a dialyzer failure that should be addressed.

src/handlers/miner_dkg_handler.erl
```diff
@@ -242,34 +242,125 @@ callback_message(Actor, Message, _State) ->
...
+           delay => Delay},
+    M = maps:map(fun(_K, Term) -> term_to_binary(Term) end, M0),
+    maps:merge(PreSer, M).
 
-deserialize(BinState) ->
-    State = binary_to_term(BinState),
-    %% get the fun used to sign things with our swarm key
+deserialize(MapState0) when is_map(MapState0) ->
+    MapState = maps:map(fun(g1, B) ->
+                                B;
```
Vagabond: perhaps call `erlang_pbc:binary_to_element` here for g1 and g2?

Vagabond: I guess we don't have curve here yet, bleh

src/miner_consensus_mgr.erl
```diff
@@ -521,85 +524,74 @@ handle_info(timeout, State) ->
...
+                                                             libp2p_group_relcast, GroupArg),
+                        Round = blockchain_block:hbbft_round(HeadBlock),
+                        case wait_for_group(Group) of
+                            started ->
+                                activate_hbbft(Group, undefined, Round),
+                                State#state{active_group = Group};
+                            {error, cannot_start} ->
+                                lager:info("didn't restore consensus group, missing"),
+                                ok = libp2p_swarm:remove_group(blockchain_swarm:swarm(), Name),
+                                case BlockHeight < (ElectionHeight+ElectionDelay+10) of
```
Vagabond: Why 10?

Vagabond: oh, this is old code

PR #243, opened 2019-10-30
LongFi proto update

lthiery:
This is going to GA right now. Merge now?


lthiery:
@macpie @evanmcc @Vagabond this is in GA now

PR #245, opened 2019-11-01
validation sidecar for hbbft groups to prevent blocking

evanmcc:
makes the tests wobbly, needs https://github.com/helium/blockchain-core/pull/285 and likely some more work


src/handlers/miner_hbbft_handler.erl
```diff
@@ -94,23 +94,28 @@ handle_command({skip, Ref, Worker}, State) ->
...
             %% we've advanced to a new round, we need to destroy the old ledger context
             %% (with the old speculatively absorbed changes that may now be invalid/stale)
             %% and create a new one, and then use that new context to filter the pending
             %% transactions to remove any that have become invalid
-            Ledger0 = blockchain:ledger(Chain),
-            Ledger1 = blockchain_ledger_v1:new_context(blockchain_ledger_v1:delete_context(Ledger0)),
-            NewChain = blockchain:ledger(Ledger1, Chain),
-            lager:info("Advancing from PreviousRound: ~p to NextRound ~p and emptying hbbft buffer", [PrevRound, NextRound]),
-            case hbbft:next_round(filter_txn_buf(HBBFT, NewChain), NextRound, TxnsToRemove) of
+            Buf = hbbft:buf(HBBFT),
```
Vagabond: Should we *empty* the hbbft buffer here while we reprocess it?

src/miner_hbbft_sidecar.erl
```diff
@@ -0,0 +1,276 @@
...
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+start_link() ->
+    gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
+
+submit(Txn) ->
+    lager:info("submitting txn"),
```
Vagabond: Remove this chatter?

Vagabond:
Unit tests look like there might be some legitimate failures in there?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -262,8 +219,11 @@ handle_message(BinMsg, Index, State=#state{hbbft = HBBFT}) ->
                             lager:info("block creation for round ~p took: ~p ms", [NewRound, Duration]),
                             BinTxnsToRemove = [blockchain_txn:serialize(T) || T <- TxnsToRemove],
                             NewerHBBFT = hbbft:finalize_round(NewHBBFT, BinTxnsToRemove),
+                            Buf = hbbft:buf(NewerHBBFT),
```
Vagabond: I wonder if this is safe; not all members of the consensus group are guaranteed to reach this point before a new block comes out.

src/miner_hbbft_sidecar.erl
```diff
@@ -0,0 +1,276 @@
...
+    {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+filter_txn_buffer(Buf, Chain) ->
+    lists:filter(fun(BinTxn) ->
+                         Txn = blockchain_txn:deserialize(BinTxn),
+                         case blockchain_txn:is_valid(Txn, Chain) of
```
Vagabond: if these are "slow" transactions, can we assume their absorbs will fail because of the changes we made in core and avoid calling the validate?

Vagabond: And if the above is true, maybe we can go back to doing filtering inline in the hbbft process so we avoid potentially losing transactions because of a crash because relcast will save our buffer again

src/miner_hbbft_sidecar.erl
```diff
@@ -0,0 +1,276 @@
...
+                _ ->
+                    {Attempt, V} = start_validation(Txn, From, Chain),
+                    {noreply, State#state{validations = Validations#{Attempt => V}}}
+            end;
+        false ->
+            case blockchain_txn:is_valid(Txn, Chain) of
+                ok ->
+                    case blockchain_txn:absorb(Txn, Chain) of
+                        ok ->
+                            catch libp2p_group_relcast:handle_command(Group, Txn),
```
Vagabond: this is synchronous so if the hbbft handler is calling us with a new buffer at the same time, we can deadlock here?

PR #246, opened 2019-11-01
add timings for validation and block creation

evanmcc:


Vagabond:
Can you just make sure all the timings have their unit specified?

PR #248, opened 2019-11-04
add transaction dumping for live debugging

evanmcc:
add the ability to dump failing txns on a live node to disk, disabled by default.


src/handlers/miner_hbbft_handler.erl
```diff
@@ -416,3 +418,15 @@ filter_txn_buf(HBBFT, Chain) ->
                                   end
                           end, Buf),
     hbbft:buf(NewBuf, HBBFT).
+
+write_txn(Reason, Txn) ->
+    case application:get_env(miner, write_failed_txns, false) of
+        true ->
+            Name = ["/tmp/", io_lib:format("~b", [erlang:phash2(Txn)]), ".txn"],
```
Vagabond: Propose we put the height in the filename, so we know the height at which it failed to absorb.

PR #249, opened 2019-11-04
Do not allow more than one 1 witness per layer per hotspot

macpie:


Vagabond:
Once you figure out the poc dist tests, please merge this.

PR #250, opened 2019-11-07
Only re-generate key in if there was no key there at all

madninja:


src/miner_sup.erl
```diff
@@ -105,7 +105,7 @@ init(_Args) ->
                                                    %% slot to handle this case.
                                                    GetPublicKey(KS+1)
                                            end;
-                                       {error, _} ->
```
vihu: What about the catchall case?

vihu: Assuming it can still throw other errors...

madninja: Other errors will cause miner to crash.. And then restarted by the system. There's no good recovery or fallback at this stage

PR #256, opened 2019-11-19
Add support for txn bundles

vihu:
Related core PR: https://github.com/helium/blockchain-core/pull/294

### DIFF
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -415,8 +415,9 @@ handle_call({create_block, Stamps, Txns, HBBFTRound}, _From, State) ->
                 LastBlockTimestamp = blockchain_block:time(CurrentBlock),
                 BlockTime = miner_util:median([ X || {_, {X, _}} <- Stamps,
                                                      X > LastBlockTimestamp]),
-                {ValidTransactions, InvalidTransactions} =
-                    blockchain_txn:validate(SortedTransactions, Chain),
+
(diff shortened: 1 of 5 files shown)


src/miner.erl
```diff
@@ -410,8 +410,20 @@ handle_call({create_block, Stamps, Txns, HBBFTRound}, _From, State) ->
                 LastBlockTimestamp = blockchain_block:time(CurrentBlock),
                 BlockTime = miner_util:median([ X || {_, {X, _}} <- Stamps,
                                                      X > LastBlockTimestamp]),
-                {ValidTransactions, InvalidTransactions} =
-                    blockchain_txn:validate(SortedTransactions, Chain),
+
+                {BundledValidTransactions, InvalidTransactions} = blockchain_txn:validate(SortedTransactions, Chain),
+
+                ValidTransactions = lists:reverse(lists:foldl(fun(Txn, Acc) ->
+                                                                      case blockchain_txn:type(Txn) of
```
evanmcc: hm.  I kind of want this code not to live in miner.  can we add it to blockchain_txn?  maybe `blockchain_txn:flatten` or something?

evanmcc: we can call it here, I just don't want anything in miner to have to understand those sorts of details.

vihu:
> Why encode the ordering here?

It's handled in core now.

vihu:
Degrading this back to WIP since there are issues with successive txns.

PR #257, opened 2019-11-21
(WIP - do not merge ) miner test suite refactor

andymck:


madninja:
> hmm i played around with a version of the above and hadnt considered the wait_util called from assertAsync performs retries upon failure, if assertAsyncTrue above is used and asserts on fail then those retries will be effectively negated and the function will fail upon first stale response

But you could fix that assertAsyncTrue to do the retries on the full main expression like the other one already is, no?

andymck:
Closing as superceeded by PR #260.  Suggestions on this PR are noted and will be carried through to new

PR #261, opened 2019-12-03
log when we time out a validation

evanmcc:
this log line got dropped in the sidecar refactor


src/miner_hbbft_sidecar.erl
```diff
@@ -172,6 +172,7 @@ handle_info({Ref, Res}, #state{validations = Validations, chain = Chain, group =
                     deadline ->
                         erlang:exit(Pid, kill),
                         write_txn("timed out", Height, Txn),
+                        lager:warning("validation timed out for ~p", [Txn]),
```
Vagabond: pretty print the txn?

PR #263, opened 2019-12-03
Various improvements to PoC to increase the chances of getting receipts/witnesses

Vagabond:
See commit messages for details.

### DIFF
--- a/src/poc/miner_onion_server.erl
+++ b/src/poc/miner_onion_server.erl
@@ -16,6 +16,7 @@
     start_link/1,
     send/1,
     decrypt/2,
+    retry_decrypt/7,
     send_receipt/6,
     send_witness/4,
     send_to_router/2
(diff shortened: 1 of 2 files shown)


src/poc/miner_onion_server.erl
```diff
@@ -326,22 +318,53 @@ wait_until_next_block() ->
             ok
     end.
 
+wait_for_block(Fun, Count) when Count > 0 ->
+    ok = blockchain_event:add_handler(self()),
+    wait_for_block_(Fun, Count).
```
evanmcc: should we unregister after this completes?

evanmcc: it's only called in a spawn, is the event system going to unregister it automatically after it exits?

Vagabond:
> this seems reasonable; we should schedule some sort of refactoring pass for the onion server, it's hard to follow.

Agreed.

PR #266, opened 2019-12-09
don't crash loop on ledger mismatch so we can attempt repair

evanmcc:
the consensus manager shouldn't crash out if it doesn't know what to do with a restore, because this takes the whole system down.  It should just enter an idle state, so repair is possible.


evanmcc:
@madninja @Vagabond I think we should just merge this to get https://github.com/helium/blockchain-core/pull/331 in

PR #274, opened 2019-12-18
Add tests for poc-v6

vihu:
- Add tests for PoC v6
- Remove poc4 and prior PoC tests.


test/miner_poc_SUITE.erl
```diff
@@ -17,6 +17,9 @@
     poc_dist_v5_test/1,
```
evanmcc: can you disable the v4 and before tests?  I think that current and one fallback is enough.

PR #285, opened 2020-01-08
Add poc-v6 tests, comment out v4 tests

vihu:


Vagabond:
We might not want to remove the old tests because then we might accidentally break those code paths.

vihu:
> We might not want to remove the old tests because then we might accidentally break those code paths.

Fair, uncommented v4 tests

PR #301, opened 2020-01-29
Macpie/lorawan

macpie:


src/miner_lora.erl
```diff
@@ -181,48 +235,67 @@ handle(<<?PROTOCOL_2:8/integer-unsigned, Token:2/binary, ?TX_ACK:8/integer-unsig
...
-handle_packet([], _Gateway, State) ->
+-spec handle_packets(list(), gateway(), state()) -> state().
+handle_packets([], _Gateway, State) ->
     State;
-handle_packet([Packet|Tail], Gateway, State) ->
+handle_packets([Packet|Tail], Gateway, #state{pubkey_bin=PubKeyBin, sig_fun=SigFun}=State) ->
     Data = base64:decode(proplists:get_value(<<"data">>, Packet)),
     case route(Data) of
         error ->
+            % TODO: Should we do the other packets?
```
Vagabond: Doesn't it fall through after the case to the tail call to handle_packets?

macpie: You are right I missed that

PR #302, opened 2020-01-29
handle removed states read from saved statem file during init

andymck:
Handle removed and no longer supported states read from the statem file during init of miner_poc_statem

All miners which were in this state will now default to requesting state

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -114,10 +114,18 @@ init(Args) ->
         {ok, _State, #data{poc_restarts=POCRestarts}} when POCRestarts == 0 ->
             {ok, requesting, #data{base_dir=BaseDir, blockchain=Blockchain,
                                    address=Address, poc_interval=Delay, state=requesting}};
+        %% we loaded a state from the saved statem file, initialise with this if its an exported/supported state
         {ok, State, #data{poc_restarts = POCRestarts} = Data} ->
-            {ok, State, Data#data{base_dir=BaseDir, blockchain=Blockchain,
-                                  address=Address, poc_interval=Delay, state=State,
-                                  poc_restarts = POCRestarts - 1}}
+            case is_supported_state(State) of
+                true ->
+                    {ok, State, Data#data{base_dir=BaseDir, blockchain=Blockchain,
+                                          address=Address, poc_interval=Delay, state=State,
+                                          poc_restarts = POCRestarts - 1}};
+                false ->
+                    lager:debug("Loaded unsupported state ~p, ignoring and defaulting to requesting", [State]),
+                    {ok, requesting, #data{base_dir=BaseDir, blockchain=Blockchain,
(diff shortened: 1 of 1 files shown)


src/poc/miner_poc_statem.erl
```diff
@@ -110,10 +110,21 @@ init(Args) ->
...
+                                 _State =:= submitting ->
+            lager:debug("Loaded unsupported state, ignoring and defaulting to requesting", []),
+            {ok, requesting, #data{base_dir=BaseDir, blockchain=Blockchain,
+                                   address=Address, poc_interval=Delay, state=requesting}};
+
         %% we have attempted to unsuccessfully restart this POC too many times, give up and revert to default state
         {ok, _State, #data{poc_restarts=POCRestarts}} when POCRestarts == 0 ->
             {ok, requesting, #data{base_dir=BaseDir, blockchain=Blockchain,
                                    address=Address, poc_interval=Delay, state=requesting}};
+
```
madninja: Spurious whitespace?

src/poc/miner_poc_statem.erl
```diff
@@ -110,10 +110,21 @@ init(Args) ->
         {error, _} ->
             {ok, requesting, #data{base_dir=BaseDir, blockchain=Blockchain,
                                    address=Address, poc_interval=Delay, state=requesting}};
+        %% support for some states were removed but some miners may have it saved in their state file
+        %% so we will force these to default to requesting
+        {ok, _State, _Data} when _State =:= targeting;
```
madninja: Would this not be brittle if we change it again? Any way to just catch the states we support instead?

andymck: good point, will do that

andymck: @madninja now only supports exported states

src/poc/miner_poc_statem.erl
```diff
@@ -114,10 +114,18 @@ init(Args) ->
...
-                                  address=Address, poc_interval=Delay, state=State,
-                                  poc_restarts = POCRestarts - 1}}
+            case is_supported_state(State) of
+                true ->
+                    {ok, State, Data#data{base_dir=BaseDir, blockchain=Blockchain,
+                                          address=Address, poc_interval=Delay, state=State,
+                                          poc_restarts = POCRestarts - 1}};
+                false ->
+                    lager:debug("Loaded unsupported state ~p, ignoring and defaulting to requesting", [State]),
+                    {ok, requesting, #data{base_dir=BaseDir, blockchain=Blockchain,
```
madninja: Do you need to reset all the variables in the #data record?

andymck: state specific variables are set via the state enter handlers, so at this point only need to set global vars.

madninja: Ah I see.. thought you were needlessly assigning here

PR #304, opened 2020-01-29
LoRaWAN support

Vagabond:
This patchset adds support for routing LoRaWAN packets.

The specific mechanism used to make this work is as follows:

For LoRaWAN join packets, the APPEUI consists of `<<OUI:32/integer-unsigned-big, DeviceID:32/integer-unsigned-big>>`, this is equivalent to the LongFi address field.

On join, the router returns the device a `DevAddr`, which is a 32 bit identifier. Unfortunately with this constrained address space, we have to convert out 64 bit address space into a 32 bit one. This is not ideal, but we can use a trick where we give all devices the same DevAddr and figure out which device it is on the router side by checking the message integrity code against all known session keys for that OUI. This takes about 32 microseconds per key, so it's not terrible.

Thus, join packets are routed by the top half of the AppEUI and other packets are routed by the DevAddr.

As part of this work we have also switched from using concentrate to using the stock Semtech packet forwarder and its protocol: https://github.com/Lora-net/packet_forwarder/blob/master/PROTOCOL.TXT

Finally, we have also changed the channel scheme to use LoRaWAN US 915 channel group 7.


rebar.lock
```diff
@@ -69,7 +69,7 @@
   0},
  {<<"helium_proto">>,
   {git,"https://github.com/helium/proto.git",
-       {ref,"ba0e3f6a35ec9397b8f0eb3bc832dbfaa95775a9"}},
```
Vagabond: This was bumped by hand, we probably need to bump it through blockchain core properly.

test/miner_onion_SUITE.erl
```diff
@@ -2,9 +2,16 @@
 
 -include_lib("common_test/include/ct.hrl").
 -include_lib("eunit/include/eunit.hrl").
--include_lib("helium_proto/src/pb/longfi_pb.hrl").
 -include("miner_ct_macros.hrl").
 
+-define(PROTOCOL_2, 2).
```
macpie: Maybe put those in a `lora.hrl`?

src/poc/miner_onion_server.erl
```diff
@@ -424,49 +311,21 @@ try_decrypt(IV, OnionCompactKey, OnionKeyHash, Tag, CipherText, ECDHFun, Chain)
...
-tx_params(Len, false) when Len < 115 ->
-    {'SF8', 'CR4_6'};
-tx_params(Len, false) when Len < 139 ->
-    {'SF8', 'CR4_5'};
-tx_params(Len, false) when Len < 160 ->
-    {'SF7', 'CR4_8'};
-tx_params(_, false) ->
-    %% onion packets won't be this big, but this will top out around 180 bytes
-    {'SF7', 'CR4_7'}.
+% -spec tx_params(integer(), Fragment :: boolean()) -> {atom(), atom()}.
```
evanmcc: clean this up?

src/poc/miner_onion_server.erl
```diff
@@ -368,31 +265,21 @@ decrypt(Type, IV, OnionCompactKey, Tag, CipherText, RSSI, Stream, #state{ecdh_fu
...
-            Packet = longfi_pb:encode_msg(Req),
-            erlang:spawn(
-                fun() ->
-                    ?MODULE:send_receipt(Data, OnionCompactKey, Type, os:system_time(nanosecond), RSSI, Stream),
-                    timer:sleep(rand:uniform(?TX_RAND_SLEEP) + ?TX_MIN_SLEEP),
-                    _ = gen_udp:send(Socket, IP, Port, Packet)
-                end
-            ),
-            State#state{packet_id=((ID+1) band 16#ffffffff), pending_transmits=[{ID, Ref}|State#state.pending_transmits]};
+            %Fragmentation = application:get_env(miner, poc_fragmentation, true),
```
evanmcc: cleanup?

src/miner_lora.erl
```diff
@@ -0,0 +1,412 @@
...
+    handle_cast/2,
+    handle_info/2,
+    terminate/2,
+    code_change/3
+]).
+
+-include_lib("helium_proto/src/pb/blockchain_state_channel_v1_pb.hrl").
+-include("lora.hrl").
+
+-define(JOIN_REQUEST, 2#000).
```
evanmcc: should these be in the `lora.hrl`?

src/miner_lora.erl
```diff
@@ -0,0 +1,412 @@
...
+terminate(_Reason, #state{socket=Socket}) ->
+    gen_udp:close(Socket),
+    ok.
+
+%% ------------------------------------------------------------------
+%% Internal Function Definitions
+%% ------------------------------------------------------------------
+
+-spec mk_token(map()) -> binary().
+mk_token(Timers) ->
```
evanmcc: how long do these live?  are they important for protocol reasons?  I suppose it's not super likely that we'll issue two in a row with the same token on a restart, but it worries me a little.

also remember that `strong_rand_bytes` takes a lock in the crypto runtime, so if performance will be at all sensitive, you should use a cheaper form of randomness.

src/miner_lora.erl
```diff
@@ -0,0 +1,412 @@
...
+-spec route(binary()) -> any().
+route(Pkt) ->
+    case longfi:deserialize(Pkt) of
+        error ->
+            route_non_longfi(Pkt);
+        {ok, LongFiPkt} ->
+            %% hello longfi, my old friend
+            case longfi:type(LongFiPkt) == monolithic andalso longfi:oui(LongFiPkt) == 0 andalso longfi:device_id(LongFiPkt) == 1 of
+                true ->
+                    %<<IV:2/binary,
```
evanmcc: cleanup?

src/miner_lora.erl
```diff
@@ -0,0 +1,399 @@
...
+    handle_info/2,
+    terminate/2,
+    code_change/3
+]).
+
+-include_lib("helium_proto/src/pb/blockchain_state_channel_v1_pb.hrl").
+-include("lora.hrl").
+
+-record(gateway, {
+    mac,
```
vihu: specs?

src/miner_lora.erl
```diff
@@ -0,0 +1,399 @@
...
+-spec send_poc(binary(), any(), float(), iolist(), any()) -> ok | {error, any()}.
+send_poc(Payload, When, Freq, DataRate, Power) ->
+    gen_server:call(?MODULE, {send, Payload, When, Freq, DataRate, Power, false}, 11000).
+
+%% ------------------------------------------------------------------
+%% gen_server Function Definitions
+%% ------------------------------------------------------------------
+
+init(Args) ->
+    UDPIP = maps:get(radio_udp_bind_ip, Args),
```
vihu: Expecting that these keys are always there or the miner_lora process blows up I presume...

PR #308, opened 2020-01-31
enable targetting to continue in the event of a crash 

andymck:
Also fixes bug with poc_starts counter not being written to state file


andymck:
hmm added wrong file

PR #309, opened 2020-01-31
enable targetting to continue in the event of a crash

andymck:
Also fix bug with poc_restarts counter not being written to state file


src/poc/miner_poc_statem.erl
```diff
@@ -360,6 +345,31 @@ handle_event(_EventType, _EventContent, Data) ->
...
+    case find_request(BlockHash, Data) of
+        {ok, Block} ->
+            Height = blockchain_block:height(Block),
+            %% TODO discuss making this delay verifiable so you can't punish a hotspot by
+            %% intentionally fast-challenging them before they have the block
+            %% Is this timer necessary ??
+            timer:sleep(?BLOCK_PROPOGATION_TIME),
+            %% save the block hash in which we found the poc request, if we crash out in targeting we can pick back up
+            %% this will be nulled if we revert back to requesting or go forward to receiving
+            save_data(Data#data{targeting_data={targeting, BlockHash, PinnedLedger}}),
```
Vagabond: Shouldn't we pass the Data with the updated targeting_data through to handle_targeting?

andymck: you are right, thats an oversight.  Will fix

src/poc/miner_poc_statem.erl
```diff
@@ -408,13 +422,11 @@ handle_targeting(Entropy, Height, Ledger, Data) ->
                 {ok, V} when V < 8 ->
                     {ok, TargetPubkeyBin} = blockchain_poc_target_v2:target_v2(Entropy, Ledger, Vars),
                     lager:info("poc_v~p target found ~p, challenging, hash: ~p", [V, TargetPubkeyBin, Entropy]),
-                    self() ! {challenge, Entropy, TargetPubkeyBin, ignored, Height, Ledger, Vars},
                     handle_challenging({Entropy, ignored}, TargetPubkeyBin, ignored, Height, Ledger, Vars, Data#data{challengees=[]});
                 {ok, V} ->
-                    {ok, {TargetPubkeyBin, TargetRandState}} = blockchain_poc_target_v3:target(ChallengerAddr, Entropy, Ledger, Vars),
+                    {ok, {TargetPubkeyBin, TargetRandState}}  = blockchain_poc_target_v3:target(ChallengerAddr, Entropy, Ledger, Vars),
```
evanmcc: extra space here

src/poc/miner_poc_statem.erl
```diff
@@ -536,39 +552,43 @@ load_data(BaseDir) ->
...
+                    challengees := Challengees,
+                    packet_hashes := PacketHashes,
+                    responses := Responses,
+                    receiving_timeout := RecTimeout,
+                    poc_hash := PoCHash,
+                    mining_timeout := MiningTimeout,
+                    retry := Retry,
+                    receipts_timeout := ReceiptTimeout,
+                    poc_restarts := PoCRestarts,
+                    targeting_data := TargetingData} ->
```
evanmcc: when we upgrade here, `targeting_data` won't be present, is the intention that we're just going to crash until we give up just this once?  it's a fine strategy, I just want to be explicit about it.

src/poc/miner_poc_statem.erl
```diff
@@ -360,6 +349,31 @@ handle_event(_EventType, _EventContent, Data) ->
...
 
+
+handle_mining(BlockHash, PinnedLedger, #data{blockchain = _Chain, address = Challenger,
+                                             mining_timeout = MiningTimeout, secret = Secret} = Data)->
+    case find_request(BlockHash, Data) of
+        {ok, Block} ->
+            Height = blockchain_block:height(Block),
+            %% TODO discuss making this delay verifiable so you can't punish a hotspot by
+            %% intentionally fast-challenging them before they have the block
+            %% Is this timer necessary ??
```
evanmcc: yes.  we intentionally delay a bit because if the block has not propagated (i.e. we're on the rising edge of the gossip wave), the remote node might not have seen the block yet, and won't be able to decode the challenge, and will fail.  It's always possible, even with the delay, but we do this to give the other end a chance.

src/poc/miner_poc_statem.erl
```diff
@@ -180,44 +183,30 @@ requesting(EventType, EventContent, Data) ->
...
 %% if we dont see our POC request txn having been mined within the Mining Timeout period ( 5 blocks )
 %% then we give up on the request and transition back to requesting state
+%% we also look out for possible restarts which took place whilst handling targeting
+%% if targeting data is set then we reload the specified block using the stored hash
+%% and retry the targeting
 %%
+mining(enter, _State, #data{targeting_data = {targeting, BlockHash, PinnedLedger}} = Data)->
+    %% sorry, have to send msg via self here as state enters cannot insert events..bah...
+    %% so I either send it here or during init..feels better here
+    self ! {retry_targeting, BlockHash, PinnedLedger},
```
evanmcc: I'm unclear about the logic here.  can you spell out where this gets cleared, and maybe what the flow looks like for failure?  it looks like we might lose a tick of the timer at each state transition?

andymck: So this state enter will only be hit if we transition to mining state AND if the state field targeting_data is set.  This will be the case if we crash out whilst handling targeting.  As such this state enter is only going to be hit if we restarted the POC statem, reloaded our state and that state had targeting data set.

The targeting data will then be cleared if we subsequently move to receiving state as part of handle_challenging or if we move too requesting state from either handle_challenging or from handle_targeting itself.  So basically it will be cleared if we succeed in targeting or challenging or if either of those fail.

src/poc/miner_poc_statem.erl
andymck: gotcha

src/poc/miner_poc_statem.erl
andymck: correct, the load_data call will crash and subsequently in the init we will default too requesting state.  If you think this is too blunt I can pull each field individually from the loaded term, avoid the crash and default values for missing fields but it would add additional maintenance overhead that prob is not warranted.

src/poc/miner_poc_statem.erl
andymck: you have eyes like a hawk :) fixed

src/poc/miner_poc_statem.erl
```diff
@@ -74,7 +74,8 @@
     mining_timeout = ?MINING_TIMEOUT :: non_neg_integer(),
     retry = ?CHALLENGE_RETRY :: non_neg_integer(),
     receipts_timeout = ?RECEIPTS_TIMEOUT :: non_neg_integer(),
-    poc_restarts = ?POC_RESTARTS :: non_neg_integer()
+    poc_restarts = ?POC_RESTARTS :: non_neg_integer(),
+    targeting_data :: {atom(), blockchain_block:hash(), binary()} | undefined
```
Vagabond: Isn't the third parameter here a rocksdb snapshot? This isn't safe to serialize to disk as it will not point to anything after a node restart...?

Vagabond: I think you need to store the block height of the ledger here, so you can use ledger_at to obtain a ledger snapshot when deserializing the stored state

andymck: Well spotted! now fixed...

PR #312, opened 2020-02-01
add adjustable starting point

evanmcc:


config/dev.config
```diff
@@ -16,7 +16,7 @@
    {seed_nodes_dns, ""},
    {peerbook_update_interval, 60000},
    {peerbook_allow_rfc1918, true},
-   {peer_cache_timeout, 200}
+   {peer_cache_timeout, 20000}
```
Vagabond: Why was this changed?

evanmcc: I changed it on the suspicion that it was hosing my half-down local cluster, and it didn't seem to hurt anything?  I'll revert and test the scenario again.  it didn't seem to happen quite as bad this time but they're definitely using more CPU than I would expect.

PR #313, opened 2020-02-03
Save data upon state enters

andymck:


Vagabond:
Are we relying on the fact that we don't alter the Data except when transitioning between states?

andymck:
Data is always saved when transitioning state but we dont rely on altering/saving Data only at that point.  There are a number of cases where we are altering and saving Data without transitioning state

PR #316, opened 2020-02-05
notify the miner that we're out

evanmcc:
this seems to have gotten lost in my latest refactor of this stuff.  With this, we'll bad_sup instead of properly signaling that we're not in the group anymore.  This should remove some confusion as to when we're getting a bad_sup for other reasons, and when we've just exited the group.


Vagabond:
lgtm

PR #319, opened 2020-02-07
Attempt to fast-forward new consensus members after election appears

Vagabond:
To prevent slow block gossip holding up a new round, try to fastforward
all the new group members to the same block as we have.

### DIFF
--- a/src/miner_consensus_mgr.erl
+++ b/src/miner_consensus_mgr.erl
@@ -474,6 +474,16 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
                                                                              ElectionDelay, State2#state.chain),
                                                     HGrp
                                             end,
+                                        %% make sure the new members are on the latest block so they know to start the round now
+                                        {ok, NewConsensusAddrs} = blockchain_ledger_v1:consensus_members(blockchain:ledger(State#state.chain)),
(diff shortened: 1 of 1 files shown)


src/miner_consensus_mgr.erl
```diff
@@ -474,6 +474,16 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
                                                                              ElectionDelay, State2#state.chain),
                                                     HGrp
                                             end,
+                                        %% make sure the new members are on the latest block so they know to start the round now
+                                        {ok, NewConsensusAddrs} = blockchain_ledger_v1:consensus_members(blockchain:ledger(State#state.chain)),
+                                        {ok, OldLedger} = blockchain:ledger_at(BlockHeight - 1, State#state.chain),
```
evanmcc: would be more efficient to subtract the minimum length for a group so we don't have to play it forward so far?  or, probably better, you can get the precise height of the block from the election_info stuff, and just pull that block and get the group from that instead of replaying the ledger.

Vagabond: we'll almost certainly have a snapshot for the last block, so this should be pretty efficent in 99% of cases?

Vagabond: I'm going to merge this for now and we can revisit the performance aspect later.

PR #326, opened 2020-02-18
don't crash manager on busy dkg, don't die quite so easily

evanmcc:
there are probably a few more things to work on, but this should narrow it down and take care of the most common ones.


Vagabond:
LGTM, is the goal of raising the intensity to reduce the key loss incidence rate?

PR #328, opened 2020-02-19
seed node target

evanmcc:
This branch brings in a number of optimizations specific to the seed nodes that haven't hit upstream yet.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -52,97 +52,74 @@
     ]}
 ]}.
 
+{relx, [
+        {release, {'miner', "0.1.0"},
+         [miner, observer, runtime_tools, tools,wx, recon]},
+        {sys_config, "./config/dev.config"},
+        {overlay,
+         [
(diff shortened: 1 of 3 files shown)


vihu:
Maybe add a make target as well while at it?

rebar.config
```diff
@@ -52,97 +52,74 @@
...
+          {txn, "extensions/txn"},
+          {dkg, "extensions/dkg"},
+          {authorize, "extensions/authorize"},
+          {repair, "extensions/repair"},
+          {print_keys, "extensions/print_keys"}
+         ]},
+        {dev_mode, true},
+        {include_erts, false}
+       ]}.
+
```
fvasquez: I see that you hoisted this `relx` out of the `dev` and `aws` profiles to avoid duplication. You kept the `print_keys` extension which is the most recent change to `rebar.config` that I care about for the hotspots.

PR #330, opened 2020-02-22
Add print_keys script

madninja:
This adds a miner/bin/print_keys shell script that prints out a list of terms for onboarding/pub keys. 

To get this to work the miner_sup startup for key retrieval was moved into a new miner_keys module

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -66,6 +66,7 @@
                              {copy, "scripts/extensions/info", "bin/extensions/info"},
                              {copy, "scripts/extensions/dkg", "bin/extensions/dkg"},
                              {copy, "scripts/extensions/authorize", "bin/extensions/authorize"},
+                             {template, "scripts/extensions/print_keys", "bin/extensions/print_keys"},
                              {copy, "./_build/default/lib/blockchain/scripts/extensions/peer", "bin/extensions/peer"},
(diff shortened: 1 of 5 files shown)


scripts/print_keys
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+erl -pa {{lib_dirs}}/**/ebin -noshell -eval 'miner_keys:print_keys("{{sys_config}}")' -eval 'init:stop()'
```
Vagabond: Why not use a miner subcommand that doesn't do a rpc call?

madninja: Because I was lazy and this was quicker?

madninja: I’ll get that fixed up

public_keys.config
```diff
@@ -0,0 +1,2 @@
+{pubkey,"112YR4mqBrc6DxrVonKsYE6bb6TGab3K1vVfpF8yAebRSd5YgZDy"}.
```
Vagabond: What is this file for?

madninja: accidental checkin

Vagabond:
@fvasquez we should call this in the runit init script in nextgate and write the output to a file

PR #332, opened 2020-02-27
Support multi payment txns

vihu:
Testing `blockchain_txn_payment_v2`

Core: https://github.com/helium/blockchain-core/pull/382


vihu:
> adjust the lockfile and this should be ready to go I think

lockfile looks right no? I checked core and proto master, both appear to be at the correct sha and merged up.

PR #333, opened 2020-03-03
Use runtime path substitution in print_keys

fvasquez:
```
# miner print_keys
{pubkey,"11gaoPJfSvyQubysyG73cLcTWCesM7vK3vWvowgNXtJqUevBFp"}.
{onboarding_key,"11RbC5hvgDYW6HwcHNjH8fsxqMYmZzB2wsr6QNMo4ATZWAs94jC"}.
{animal_name,"alert-berry-trout"}.
```


scripts/extensions/print_keys
```diff
@@ -1,3 +1,11 @@
 #!/bin/sh
 
-erl -pa {{lib_dirs}}/**/ebin -noshell -eval 'miner_keys:print_keys("{{sys_config}}")' -eval 'init:stop()'
+CONFIG=$(find $ROOTDIR/releases -name sys.config)
+LIB_DIR=$ROOTDIR/lib
+
+if relx_nodetool "ping" > /dev/null 2>&1; then
+    relx_nodetool "rpcterms" miner_keys print_keys "'${CONFIG}'." | sed '$d'
```
Vagabond: What does `sed '$d'` do?

Vagabond: I don't think you want to call this same function in both contexts, if the miner is running there's already a process managing the ecc508, this call will boot a new one.

fvasquez: It strips out the trailing `ok` line from the rpc result.

fvasquez: > I don't think you want to call this same function in both contexts, if the miner is running there's already a process managing the ecc508, this call will boot a new one.

You are correct.  From the comments for the `keys` fucntion:

```
%% NOTE: Do NOT call this after miner has started since this function
%% will attempt to communicate directly with the ECC. Use only as part
%% of startup or other miner-free scripts.
```

So unless miner has the equivalent of `dump_keys` defined somewhere I should just revert back to `cat /var/data/public_keys` when miner is running?

fvasquez: Reverted back to `cat /var/data/public_keys`.

scripts/extensions/print_keys
```diff
@@ -1,3 +1,11 @@
 #!/bin/sh
 
-erl -pa {{lib_dirs}}/**/ebin -noshell -eval 'miner_keys:print_keys("{{sys_config}}")' -eval 'init:stop()'
+CONFIG=$(find $ROOTDIR/releases -name sys.config)
+LIB_DIR=$ROOTDIR/lib
+
+if relx_nodetool "ping" > /dev/null 2>&1; then
+    cat /var/data/public_keys
```
Vagabond: do an exec here, so the exit code will propogate if the file doesn't exist?

Vagabond: Although you can almost certainly get the information you want out of the running swarm??

fvasquez: > do an exec here, so the exit code will propogate if the file doesn't exist?

`exec` will also terminate the `print_keys` script so that `exit 0` is never reached and thereby superfluous?

fvasquez: No more cat now that #335 has been merged into this PR.

Vagabond:
The # was needed to stop it deleting the last line, I meant to remove the sed entirely

fvasquez:
> The # was needed to stop it deleting the last line, I meant to remove the sed entirely

If miner and gateway_config have no problems parsing the last `ok` line from `/var/data/public_keys` then I'm fine with removing the sed.

Vagabond:
It wasn't putting the ok in there anymore, AFAICT

Vagabond:
but it should be tested

PR #336, opened 2020-03-06
pad single digit ints when displaying mac address

andymck:


src/cli/miner_cli_info.erl
```diff
@@ -362,5 +362,10 @@ format_macs_from_interfaces(IFs)->
     [], IFs).
 
 format_hwaddr(HWAddr)->
-    HWAddr0 = [integer_to_list(B,16) || B <- HWAddr ],
+    HWAddr0 = [pad_and_format_int(B) || B <- HWAddr ],
```
Vagabond: Worth looking at `blockchain_utils:bin_to_hex/1` instead?

Vagabond:
```suggestion
   blockchain_utils:bin_to_hex(list_to_binary(HWAddr)).
```

andymck: looks good.  I tested your suggestion locally and does the job

Vagabond:
I think the only annoying thing is that it outputs lowercase hex, maybe check if we want uppercase?

andymck:
will check with redbeard, also need to remove a now spurious list:concat

PR #337, opened 2020-03-06
Fix various issues with the miner tests

Vagabond:
* Reduce unnecessary logging noise
* Ensure gossip networks get initialized in multi node tests
* Various other tweaks to improve test resilience


test/miner_poc_SUITE.erl
```diff
@@ -670,7 +670,7 @@ setup_dist_test(TestCase, Config, VarMap) ->
     true = load_genesis_block(GenesisBlock, Miners, Config),
     miner_fake_radio_backplane ! go,
     %% wait till height 50
-    true = wait_until_height(Miners, 50),
+    true = wait_until_height(Miners, 10),
```
vihu: maybe we should also lower the `poc_challenge_interval` for testing from 20 -> 10?

Vagabond: Tried this, it breaks the test

PR #339, opened 2020-03-09
Test state channels

vihu:
Add tests for data credits


vihu:
#346 supersedes this one

PR #340, opened 2020-03-09
thread seen and BBA completion state into the block

evanmcc:
We'd like to be able to eject poorly performing members before they cause block time problems.  This code allows us to thread information through into the block so that we can deterministically remove nodes without waiting for them to become inactive, which scales poorly with lowering the PoC request rate.

TODO: on the blockchain side, elections need to be rewritten to take this information into account.  We also should consider gating this behind a chain var, but I dunno should work (usage of this information definitely goes behind a var)?
 
requires:

https://github.com/helium/blockchain-core/pull/392
https://github.com/helium/proto/pull/21


Vagabond:
I think we need a chain var to enable this extra information, otherwise we might produce a a block that cannot be signed because one of the nodes can't understand the block metadata? Or would that clear once the lagging node upgraded?

PR #342, opened 2020-03-11
Attempt to upload logs on test failures

Vagabond:


andymck:
Confirmed this is working and very helpful.  The only point to note is that the tar files can be somewhat large ( 300mb ).  Maybe we can exclude the data folder to reduce the size?  I dont think the data there will be overly helpful to people?

Vagabond:
I think it's worth investigating what's in there that we're not culling from successful tests, for some of the tests having the data files might be very helpful.

andymck:
@Vagabond the private data dir for each test account for the vast bulk of the exploded tar filesize. A 300MB tar file unzips to 1.47GB, of this the private data dirs account for 1.32GB

In a miner test run, it produces ~280 data dirs, these range in size between 1MB - 8MB and consist of the following subfolders:

blockchain.db
ledger.db
blockchain_swarm

Atm the tar file contains the above for all test cases, whether they are successful or not.  Miner does attempt to cleanup the logs of successful test runs but on a review I can see the target dirs it is attempting to clean are incorrectly defined.  I'll fix this, should bring the filesize down dramatically for most runs

Vagabond:
How do you think we should confirm this is working ?

andymck:
I actually tested this on a previous PR ( copied the changes over ).  Byt you can also see it working in the last buildkite run of this PR. 

I have downloaded the artifacts and confirmed they work, you can browse the CT results via index.html ( all links are working ).  I can also confirm the only remaining private data is that associated with failed tests.  Priv data for all successful tests has been deleted.

Vagabond:
Yeah, I guess as long as we continue trying to get the tests to be more reliable, we won't fill up S3 too quickly (presumably we could do periodic deletes?) Maybe we can make the artifacts overwrite each other? (keep at most one per branch?)

andymck:
One artifact per branch seems like a good idea.  I can take a look at that

Vagabond:
One of the buildkite env vars is the branch name, IIRC, so maybe if we created the tarball with the branch name instead of the commit sha or whatever it'd replace the file on S3

andymck:
so there is a prob with using the branch name.  During a buildkite build the head is detached and any regular means ( ie buildkite env var or shell command ) to get the branch name fails ( in that it only returns HEAD ).
I have attempted some trickery which does work on my local but for whatever reason wont work on buildkite ( always returns an empty string ).  @Vagabond check over my latest changes and see if you can see any issue or if you know a better solution

Vagabond:
it looks like it DID work on this build: https://buildkite.com/helium/miner/builds/1576#c41a8191-33b3-4cd5-93d8-42cee5aa3ce7

Makefile
```diff
@@ -20,7 +26,16 @@ typecheck:
 	$(REBAR) dialyzer xref
 
 ci: compile
-	$(REBAR) do dialyzer,xref && $(REBAR) do eunit,ct
+	($(REBAR) do eunit,ct --suite test/miner_blockchain_SUITE || (mkdir -p artifacts; tar --exclude='./_build/test/lib' --exclude='./_build/test/plugins' -czf artifacts/$(CIBRANCH).tar.gz _build/test; false))
+
+release:
```
Vagabond, line 31: How did these get duplicated?

Makefile
```diff
@@ -20,7 +26,16 @@ typecheck:
 	$(REBAR) dialyzer xref
 
 ci: compile
-	$(REBAR) do dialyzer,xref && $(REBAR) do eunit,ct
+	($(REBAR) do eunit,ct --suite test/miner_blockchain_SUITE || (mkdir -p artifacts; tar --exclude='./_build/test/lib' --exclude='./_build/test/plugins' -czf artifacts/$(CIBRANCH).tar.gz _build/test; false))
```
Vagabond, line 29: Obviously fix this to run the tests again properly and do xref/dialyzer as well

Makefile
andymck: no idea, guess i was trigger happy at some point...will clean up

andymck:
Strange its working with the branch name, maybe i was mistakenly looking at an older build originally

andymck:
branch is cleaned up and good to go

PR #343, opened 2020-03-13
Tweak the behaviour of miner_lora a bit

Vagabond:
* Use reuseaddr so we have a better chance of grabbing the port after a
  crash, or during tests
* Assume any non onion packet is a lorawan packet, we've seen some false
  positives

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -5,7 +5,8 @@
 -export([
     start_link/1,
     send/1,
-    send_poc/5
+    send_poc/5,
+    port/0
 ]).
 
 -export([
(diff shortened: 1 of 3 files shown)


src/miner_lora.erl
```diff
@@ -59,14 +60,18 @@ send(#packet_pb{payload=Payload, timestamp=When, signal_strength=Power, frequenc
 send_poc(Payload, When, Freq, DataRate, Power) ->
     gen_server:call(?MODULE, {send, Payload, When, Freq, DataRate, Power, false}, 11000).
 
+-spec port() -> {ok, inet:port_number()} | {error, any()}.
+port() ->
```
vihu: Is this fun only here for debugging purposes? I don't see anything invoking it...

Vagabond: look harder

Vagabond: https://github.com/helium/miner/pull/343/files#diff-0a89925ece2208282952849cb72c97f2R663

vihu: Ah yes :+1:

PR #344, opened 2020-03-13
Fix potential instability in POC test suite 

andymck:
the POC suite has a tendency to fail randomly.  I traced this back to the onion server crashing out a number of times in quick succession and breaching the miner_sup restart intensity leading to the entire miner tree restarting.

Each poc suite test initialises with the the following:

{_, Locations} = lists:unzip(initialize_chain(Miners, TestCase, Config, VarMap)),
GenesisBlock = get_genesis_block(Miners, Config),
miner_fake_radio_backplane:start_link(maps:get(?poc_version, VarMap), 45000, lists:zip(lists:seq(46001, 46000 + MinerCount), Locations)),
timer:sleep(5000),
true = load_genesis_block(GenesisBlock, Miners, Config),

As part of `initialize_chain` there is a call to the consensus mgr to do the initial_dkg.   For those nodes part of the CG group this takes sufficient time to complete to allow the chain to be fully setup before we advance to the main part of the tests.

However for the node not part of the CG group and thus not part of the DKG, the call to consensus mgr returns much quicker and we move onto the main part of the tests sooner.  At this point the onion server may receive decrypt requests which will then result in calls to blockchain_worker to get the chain, followed by a call to the ledger.   At times the blockchain worker will not be fully setup with the chain and it will return undefined.  At this point and the onion server will crash out on `Ledger = blockchain:ledger(Chain)` calls

The miner sup will then immediately restart the onion server and the same crash will re-occur.  As the miner sup restart strategy only allows 2 restarts in 5 seconds we then see the entire miner tree being restarted and the tests will fail.

To address the above, this fix simply captures and drops all decrypt requests whilst the there is no chain and re attempts to get the chain from blockchain worker.

### DIFF
--- a/src/miner_sup.erl
+++ b/src/miner_sup.erl
@@ -42,7 +42,7 @@ start_link() ->
 init(_Args) ->
     SupFlags = #{
         strategy => rest_for_one,
-        intensity => 2,
+        intensity => 10,
         period => 5
     },
 
--- a/src/miner_sup.erl
+++ b/src/miner_sup.erl
@@ -42,7 +42,7 @@ start_link() ->
 init(_Args) ->
     SupFlags = #{
         strategy => rest_for_one,
-        intensity => 10,
+        intensity => 2,
         period => 5
     },
 
--- a/src/poc/miner_onion_server.erl
+++ b/src/poc/miner_onion_server.erl
@@ -50,8 +50,7 @@
     miner_name :: binary(),
     sender :: undefined | {pid(), term()},
     packet_id = 0 :: non_neg_integer(),
-    chain :: undefined  | blockchain:blockchain(),
-    chain_retry_count = 0 :: integer()
+    chain :: undefined  | blockchain:blockchain()
 }).
 
 -define(BLOCK_RETRY_COUNT, 10).
--- a/src/poc/miner_onion_server.erl
+++ b/src/poc/miner_onion_server.erl
@@ -49,7 +49,9 @@
     ecdh_fun,
     miner_name :: binary(),
     sender :: undefined | {pid(), term()},
-    packet_id = 0 :: non_neg_integer()
+    packet_id = 0 :: non_neg_integer(),
+    chain :: undefined  | blockchain:blockchain(),
+    chain_retry_count = 0 :: integer()
 }).
 
 -define(BLOCK_RETRY_COUNT, 10).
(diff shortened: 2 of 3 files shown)


src/poc/miner_onion_server.erl
```diff
@@ -232,10 +240,9 @@ wait_for_block_(Fun, Count) ->
             wait_for_block_(Fun, Count - 1)
     end.
 
-decrypt(Type, IV, OnionCompactKey, Tag, CipherText, RSSI, Stream, #state{ecdh_fun=ECDHFun}=State) ->
+decrypt(Type, IV, OnionCompactKey, Tag, CipherText, RSSI, Stream, #state{ecdh_fun=ECDHFun, chain = Chain}=State) ->
```
Vagabond: What if chain is undefined here?

andymck: if chain in undefined, we will never hit decrypt...all casts whilst chain is undefined will be caught by the new handler

src/miner_sup.erl
```diff
@@ -42,7 +42,7 @@ start_link() ->
 init(_Args) ->
     SupFlags = #{
         strategy => rest_for_one,
-        intensity => 2,
+        intensity => 10,
```
vihu: Why was this bumped to 10?

andymck: the onion server was crashing because it was handling msgs before the chain was fully up, it was a startup timing issue.  The restart strategy here was ineffective, in that 2 did not give the much wriggle room for the chain to come back up before the next restart of the onion server and thus we hit a breach of the sup intensity and the entire tree was restarted.

Effectively with a value of 2/5 the rest_for_one strategy is no better than a one_for_all.   So this change is to give it that little bit of additional grace before we take down the entire miner tree.

Vagabond: If we've fixed the crash, does this still matter?

andymck: It doesnt matter in relation to this issue being addressed but same issue could occur in other scenarios.  I am fine with removing the intensity change if there is concern about it, but the fact that the miner tree can easily be taken down is something to be aware off

Vagabond: Miner crashing is fine, runit will restart it. I'd prefer we retract this change for now, the rest of the PR seems fine.

src/poc/miner_onion_server.erl
```diff
@@ -168,21 +169,28 @@ send_witness(Data, OnionCompactKey, Time, RSSI, Retry) ->
 %% gen_server Function Definitions
 %% ------------------------------------------------------------------
 init(Args) ->
+    lager:info("init with ~p", [Args]),
```
Vagabond: Does this add much value?

andymck: this line was already present, i just moved it up....personally i think its good to have all core servers say hello on startup.  Gives you the warm reassurance they are up and allows you to easily identify the PID when you need it...

Vagabond: Fair.

PR #349, opened 2020-03-17
Attempt to be more flexible about parsing Semtech JSON packets

Vagabond:
Additionally fix an edge case when no gateways are connected and also
fix a bug in the fake radio backplane.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -157,6 +157,9 @@ mk_token(Timers) ->
 
 -spec select_gateway(map()) -> {ok, gateway()} | {error, no_gateways}.
 select_gateway(Gateways) ->
+    %% TODO for a multi-tenant miner we'd have a mapping of swarm keys to
+    %% 64-bit packet forwarder IDs and, depending on what swarm key this send
+    %% was directed to, we'd select the appropriate gateway from the map.
     case maps:size(Gateways) of
         0 ->
             {error, no_gateways};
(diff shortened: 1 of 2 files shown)


vihu:
There are a lot of `maps:get/2` calls here which I'm assuming are meant to blow up if the key isn't found?

src/miner_lora.erl
```diff
@@ -151,9 +155,14 @@ mk_token(Timers) ->
...
 
--spec select_gateway(map()) -> gateway().
+-spec select_gateway(map()) -> {ok, gateway()} | {error, no_gateways}.
 select_gateway(Gateways) ->
-    erlang:element(2, erlang:hd(maps:to_list(Gateways))).
+    case maps:size(Gateways) of
+        0 ->
+            {error, no_gateways};
+        _ ->
+            {ok, erlang:element(2, erlang:hd(maps:to_list(Gateways)))}
```
vihu: Select the first gateway always? Can we add a comment here if we need to update this down the line?

Vagabond:
> There are a lot of `maps:get/2` calls here which I'm assuming are meant to blow up if the key isn't found?
Yes, they're mandatory fields

src/miner_lora.erl
Vagabond: yeah, this will change with the 'multi tenant' miner

Vagabond: Added a comment

Vagabond:
This was tested by a community member and it seems to have fixed his problem where his JSON had unexpected fields (because he was running a TTN packet forwarder).

Vagabond:
I didnt mean to upgrade rebar3, will revert that.

PR #350, opened 2020-03-19
Optionally mirror all packet forwarder up/down packets to a seperate udp socket

Vagabond:
This was done for @lthiery to enable the development of a LoRaWAN protocol sniffer and he has confirmed it works.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -38,7 +38,8 @@
     gateways = #{}, %% keyed by MAC
     packet_timers = #{}, %% keyed by token
     sig_fun,
-    pubkey_bin
+    pubkey_bin,
+    mirror_socket
 }).
 
 -type state() :: #state{}.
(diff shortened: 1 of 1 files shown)


src/miner_lora.erl
```diff
@@ -127,11 +137,15 @@ handle_info({tx_timeout, Token}, #state{packet_timers=Timers}=State) ->
     end,
     {noreply, State#state{packet_timers=maps:remove(Token, Timers)}};
 handle_info({udp, Socket, IP, Port, Packet}, #state{socket=Socket}=State) ->
+    maybe_mirror(State#state.mirror_socket, Packet),
```
andymck, line 140: Care about the response here from gen_udp:send ?

src/miner_lora.erl
```diff
@@ -72,8 +73,16 @@ init(Args) ->
     UDPIP = maps:get(radio_udp_bind_ip, Args),
     UDPPort = maps:get(radio_udp_bind_port, Args),
     {ok, Socket} = gen_udp:open(UDPPort, [binary, {reuseaddr, true}, {active, 100}, {ip, UDPIP}]),
+    MirrorSocket = case application:get_env(miner, radio_mirror_port, undefined) of
+        undefined ->
+            undefined;
+        P ->
+            {ok, S} = gen_udp:open(P, [binary, {active, true}]),
```
andymck, line 80: This mirror socket is not being closed in the terminate.  
In addition the process itself is not trapping exits and thus risk terminate and its cleanup not being called

src/miner_lora.erl
Vagabond: nope, fire and forget

src/miner_lora.erl
Vagabond: The socket will close when the controlling process closes, no?

andymck: That would be the expectation but with sockets i think its always good practise to explicitly close them and minimise any potential oddities that could lead to a leak

Vagabond: I think we can rely on BEAM closing the socket when the controlling process dies, that's pretty much guaranteed beyond a BEAM bug.

lthiery:
Is this landable? Having this in production miner would help in some cases debugging with third parties.

JayKickliter:
Could it be better to implement this in the forwarder itself?

PR #351, opened 2020-03-24
allow recent sync to drive poc statem forward

evanmcc:
we're seeing some issues where hotspots fall into sync mode and stop requesting for a long while.  allowing sync events to drive the statem when recent should allow them to challenge in a more timely fashion.

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -159,22 +159,34 @@ requesting(info, Msg, #data{blockchain = Chain} = Data) when Chain =:= undefined
         NewChain ->
             {keep_state, Data#data{blockchain=NewChain}, [{next_event, info, Msg}]}
     end;
-requesting(info, {blockchain_event, {add_block, BlockHash, false, Ledger}}, #data{address=Address}=Data) ->
-    case allow_request(BlockHash, Data) of
-        false ->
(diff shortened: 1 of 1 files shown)


src/poc/miner_poc_statem.erl
```diff
@@ -159,22 +159,34 @@ requesting(info, Msg, #data{blockchain = Chain} = Data) when Chain =:= undefined
...
-            lager:info("submitted poc request ~p", [Txn]),
-            {next_state, mining, save_data(Data#data{state=mining, secret=Secret,
-                                                      onion_keys=Keys, responses=#{},
-                                                      poc_hash=BlockHash})}
+requesting(info, {blockchain_event, {add_block, BlockHash, Sync, Ledger}} = Msg,
+           #data{address=Address}=Data) ->
+    Now = erlang:system_time(seconds),
+    case blockchain:get_block(BlockHash, Data#data.blockchain) of
+        {ok, Block} ->
+            case Sync andalso (Now - blockchain_block:time(Block) > 3600) of
```
Vagabond: So if this is not a sync block OR it's a sync block newer than 1 hour, we try to make progress?

evanmcc: yes.  the code comes directly from the consensus manager

Vagabond: If we need to duplicate it again, lets make a utility function.

PR #362, opened 2020-04-14
Build miner inside Docker container for aarch64

fvasquez:
The `Dockerfile` builds and runs miner inside a container using QEMU.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,13 +11,25 @@ steps:
       - "artifacts/*"
 
   - if: build.tag != null
-    name: ":whale: docker"
+    name: ":whale: ARM64 docker"
     agents:
       queue: "erlang"
     commands:
       - "git fetch -t"
       - ".buildkite/scripts/make_arm64_docker.sh"
+
+  - if: build.tag != null
+    name: ":whale: ARM32 docker"
+    agents:
+      queue: "erlang"
+    commands:
+      - "git fetch -t"
       - ".buildkite/scripts/make_arm32_docker.sh"
 
-
-
+  - if: build.tag != null
+    name: ":whale: AMD64 docker"
+    agents:
+      queue: "erlang"
+    commands:
+      - "git fetch -t"
+      - ".buildkite/scripts/make_amd64_docker.sh"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,3 +9,15 @@ steps:
       queue: "erlang"
     artifact_paths:
       - "artifacts/*"
+
+  - if: build.tag != null
+    name: ":whale: docker"
+    agents:
+      queue: "erlang"
+    commands:
+      - "git fetch -t"
+      - ".buildkite/scripts/make_arm64_docker.sh"
+      - ".buildkite/scripts/make_arm32_docker.sh"
+
+
+
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,6 +5,7 @@ steps:
       - "make compile"
       - "make ci"
     name: ":hammer: build"
+    key: "tests"
     agents:
       queue: "erlang"
     artifact_paths:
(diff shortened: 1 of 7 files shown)


mrallen1:
Looks good. Tomorrow let's work on getting this running in buildkite

mrallen1:
Despite the tests failure, I believe this PR is ready for review and merging. @Vagabond @evanmcc

.buildkite/pipeline.yml
```diff
@@ -5,7 +5,36 @@ steps:
...
+#
+#  - if: build.tag != null
+#    name: ":whale: ARM32 docker"
+#   agents:
+#     queue: "erlang"
+#   commands:
+#     - "git fetch -t"
+#     - ".buildkite/scripts/make_arm32_docker.sh"
+
+  - if: build.tag != null
```
madninja, line 34: is this section duplicated?

.buildkite/scripts/Dockerfile-amd64
```diff
@@ -0,0 +1,40 @@
...
+
+WORKDIR /opt/miner
+
+ENV COOKIE=miner \
+    # Write files generated during startup to /tmp
+    RELX_OUT_FILE_PATH=/tmp
+
+COPY --from=builder /opt/rel /opt/miner
+
+#CMD tail -f /dev/null
```
madninja, line 40: Should the docker file strip out the source or other build shrapnel after building a release?

.buildkite/scripts/Dockerfile-arm32
```diff
@@ -0,0 +1,41 @@
...
+RUN ulimit -n 64000
+
+WORKDIR /opt/miner
+
+ENV COOKIE=miner \
+    # Write files generated during startup to /tmp
+    RELX_OUT_FILE_PATH=/tmp
+
+COPY --from=builder /opt/rel /opt/miner
+
```
madninja, line 41: Same question? Does cleaning up a lot of the source and build output help make the image slimmer?

.buildkite/scripts/Dockerfile-arm64
```diff
@@ -0,0 +1,40 @@
...
+
+WORKDIR /opt/miner
+
+ENV COOKIE=miner \
+    # Write files generated during startup to /tmp
+    RELX_OUT_FILE_PATH=/tmp
+
+COPY --from=builder /opt/rel /opt/miner
+
+#CMD tail -f /dev/null
```
madninja, line 40: Same question on retaining build/ source shrapnel after the docker build

.buildkite/pipeline.yml
```diff
@@ -5,7 +5,36 @@ steps:
       - "make compile"
       - "make ci"
     name: ":hammer: build"
+    key: "tests"
```
madninja, line 8: What is this key for?

mrallen1: Well, eventually I'd like to make building the image on having a clean test run, so this just sets the stage for that. Down the road we can add a

```yaml
  depends_on: tests
```

line in there.

.buildkite/pipeline.yml
mrallen1: Yes, so the builds will run in parallel with each other and only be trigged by a tag

.buildkite/scripts/Dockerfile-amd64
mrallen1: This `COPY` command basically leaves all the build foo in a different container but moves our release into a fresh container.

.buildkite/scripts/Dockerfile-arm32
mrallen1: Yes it does

.buildkite/pipeline.yml
madninja: Right but it's building the same make_amd64_docker.sh twice.. in parallel

.buildkite/scripts/Dockerfile-arm64
mrallen1: Yeah, tbh we should probably purge the build images from bk - that's a good clean up step.

mrallen1: Honestly, we should probably just purge images in general after... 10 days maybe?

.buildkite/pipeline.yml
madninja: I can't read worth crap today.

PR #364, opened 2020-04-14
Add test for zero amount payment_v2 txns

vihu:
Related to: https://github.com/helium/blockchain-core/pull/421

### DIFF
--- a/test/miner_payment_v2_txn_SUITE.erl
+++ b/test/miner_payment_v2_txn_SUITE.erl
@@ -15,13 +15,15 @@
         ]).
 
 -export([
-         basic_test/1
+         basic_test/1,
+         zero_amt_test/1
         ]).
 
 %% common test callbacks
 
 all() -> [
-          basic_test
+          basic_test,
+          zero_amt_test
          ].
 
 init_per_suite(Config) ->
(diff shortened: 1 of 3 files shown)


test/miner_payment_v2_txn_SUITE.erl
```diff
@@ -137,3 +140,64 @@ basic_test(Config) ->
...
+    PayeeAmount = 0,
+    Payments = [blockchain_payment_v2:new(P, PayeeAmount) || P <- PayeeAddrs],
+    Txn = ct_rpc:call(Payer, blockchain_txn_payment_v2, new, [PayerAddr, Payments, 1, Fee]),
+
+    {ok, _Pubkey, SigFun, _ECDHFun} = ct_rpc:call(Payer, blockchain_swarm, keys, []),
+
+    SignedTxn = ct_rpc:call(Payer, blockchain_txn_payment_v2, sign, [Txn, SigFun]),
+    ct:pal("SignedTxn: ~p", [SignedTxn]),
+
+    {[], [SignedTxn]} = ct_rpc:call(Payer, blockchain_txn, validate, [[SignedTxn], Chain]),
```
andymck, line 183: Kinda a nit but maybe better to do something like ?assertEqual(false, TXN_TYPE:is_valid(Txn, Chain)) ?
Make the validation check more explicit and obvious

vihu: Certainly, added d149bca

PR #368, opened 2020-04-17
bump to blockchain core version with txn mgr cache as records

andymck:
Update test to account for txn mgr's txn_list returning a map of txns ( key: txn, value: TxnData )

tweak some test timers for better stability

Associated blockchain core pr: https://github.com/helium/blockchain-core/pull/427


andymck:
superceeded by https://github.com/helium/miner/pull/372

PR #375, opened 2020-04-21
Plug some holes in the hbbft signature phase

Vagabond:
Fix some places where we would incorrectly drop signatures or not
transition to the gossip or done states.


src/handlers/miner_hbbft_handler.erl
```diff
@@ -194,11 +194,14 @@ handle_message(BinMsg, Index, State=#state{hbbft = HBBFT}) ->
             Sigs = dedup_signatures(Signatures, State),
             NewState = State#state{signatures = Sigs},
             case enough_signatures(NewState) of
-                {ok, done, Signatures} when Round > NewState#state.signed ->
+                {ok, done, MySignatures} when Round > NewState#state.signed ->
```
evanmcc, line 197: ow. This is what I get for copying this around.  maybe we should make it a function?

PR #378, opened 2020-04-26
Ignore - test tweaks WIP

andymck:
This fixes the miner's blockchain_SUITE tests.  Two primary problems are addressed:

1.  The test includes a call to blockchain_gossip_handler:add_block.  This arity here was incorrect since https://github.com/helium/blockchain-core/commit/40515021e26de76534e148846c72d429e9f06080#diff-70ab3c21b9b17903889bfc6a4d66fac4

2.  Instability was introduced to the test with an incorrect use of miner_ct_util:wait_for_gte since the miner tests were refactored by me back in November.  I compared the current against the test prior to this work and made a number of corrections.

### DIFF
--- a/test/miner_blockchain_SUITE.erl
+++ b/test/miner_blockchain_SUITE.erl
@@ -223,7 +223,8 @@ election_test(Config) ->
 
     %% we've seen all of the nodes, yay.  now make sure that more than
     %% one election can happen.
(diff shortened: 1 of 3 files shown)


andymck:
superseded by https://github.com/helium/miner/pull/380

PR #379, opened 2020-04-26
Relax poc receipt submit timeouts and track transactions via callbacks

Vagabond:
If the submitting node is not more than 5 blocks behind the chain it
will never see its poc request hit the chain and thus end up in a loop
of never submitting receipts. This lengthens the poc request submission
timeout to the same as the poc interval and uses the transaction
callback to log if the receipt has been submitted.

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -43,7 +43,7 @@
 -include_lib("blockchain/include/blockchain_vars.hrl").
 
 -define(SERVER, ?MODULE).
--define(MINING_TIMEOUT, 5).
+-define(MINING_TIMEOUT, 30).
 -define(CHALLENGE_RETRY, 3).
 -define(RECEIVING_TIMEOUT, 10).
 -define(RECEIPTS_TIMEOUT, 10).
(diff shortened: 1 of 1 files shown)


src/poc/miner_poc_statem.erl
```diff
@@ -206,6 +209,9 @@ mining(enter, _State, Data)->
     {keep_state, Data};
 mining(info, Msg, #data{blockchain=undefined}=Data) ->
     handle_event(info, Msg, Data);
+mining(info, {TxnRef, Result}, #data{txn_ref=TxnRef}=Data) ->
+    lager:info("PoC request submission result ~p", [Result]),
+    {keep_state, Data};
```
vihu, line 214: If you got the result for the `txn_ref` why keep the ref in `Data` at this point?

Vagabond: Why not?

vihu: Hmm, I guess its fine since it will get a new ref on new submission anyway

Vagabond:
Yeah, we can certainly be more sophisticated here, this was just to improve behaviour in the simplest way possible

PR #385, opened 2020-05-05
move in_consensus implementation out of consensus mgr and into cli

andymck:
The check does not rely on anything within consensus_mgr and has been seen to result crashing the consensus mgr to the the call to `blockchain_swarm:pubkey_bin()` timing out ( due to sync call to retrieve the PID timing out )

As such moving implementation into the CLI itself reduces the risk to consensus mgr....


andymck:
Closing as the problem leading to the crash is addressed in another PR

PR #388, opened 2020-05-06
ledger snapshots

evanmcc:
miner support and tests for:

https://github.com/helium/blockchain-core/pull/403


src/miner.erl
```diff
@@ -367,6 +369,31 @@ handle_call({create_block, Metadata, Txns, HBBFTRound}, _From, State) ->
...
                     Metadata),
+    %% TODO what do we do about snapshot disagreements?
+    %% IIRC we had agreed to omit
+    SnapshotHash=
+        case blockchain:config(?snapshot_interval, blockchain:ledger(Chain)) of
+            {ok, Interval} ->
+                case (NewHeight - 1) rem Interval == 0 of
+                    true ->
+                        lager:info("XXX trying to add hash ~p", [NewHeight]),
+                        [SH] = lists:foldl(
```
Vagabond, line 380: badmatching here would presumably halt block production?

evanmcc: yes, I need to change this to 2f+1 agreement, and omission if it does not meet that standard.

rebar.config
```diff
@@ -11,7 +11,8 @@
 {deps, [
         {blockchain, {git, "https://github.com/helium/blockchain-core.git",
                       {branch, "master"}}},
-        {hbbft, {git, "https://github.com/helium/erlang-hbbft.git", {branch, "master"}}},
+        {hbbft, {git, "https://github.com/helium/erlang-hbbft.git",
```
Vagabond, line 14: This branch merged.

evanmcc: fixed

src/miner.erl
```diff
@@ -369,6 +371,35 @@ handle_call({create_block, Metadata, Txns, HBBFTRound}, _From, State) ->
...
+                        SHCt =
+                            lists:foldl(
+                              fun({_Idx, #{snapshot_hash := SH}}, Acc) -> % new map vsn
+                                      maps:update_with(SH, fun(V) -> V + 1 end, 1, Acc);
+                                 (_, Acc) ->
+                                      Acc
+                              end,
+                              #{},
+                              Metadata),
+                        case lists:reverse(lists:keysort(2, maps:to_list(SHCt))) of
```
Vagabond, line 394: So we're sorting the most common ones first?

Vagabond: Couple comments here would be good, I think.

src/miner.erl
```diff
@@ -369,6 +371,35 @@ handle_call({create_block, Metadata, Txns, HBBFTRound}, _From, State) ->
                     end,
                     [],
                     Metadata),
+    %% TODO what do we do about snapshot disagreements?
```
Vagabond, line 374: Update this now we have decided

src/miner.erl
```diff
@@ -414,7 +451,8 @@ handle_call({create_block, Metadata, Txns, HBBFTRound}, _From, State) ->
                                election_epoch => ElectionEpoch,
                                epoch_start => EpochStart,
                                seen_votes => SeenVectors,
-                               bba_completion => BBA
+                               bba_completion => BBA,
+                               snapshot_hash => SnapshotHash
```
Vagabond, line 455: The empty binary is the PBC default value, right?

src/miner_hbbft_sidecar.erl
```diff
@@ -74,9 +74,16 @@ init(State) ->
             erlang:send_after(500, self(), chain_check),
             {ok, State#state{}};
         Chain ->
-            Ledger = blockchain_ledger_v1:new_context(blockchain:ledger(Chain)),
-            Chain1 = blockchain:ledger(Ledger, Chain),
-            {ok, State#state{chain = Chain1}}
+            Ledger0 = blockchain:ledger(Chain),
+            case Ledger0 of
```
Vagabond, line 78: Do we still have the undefined ledger so we need this?

src/miner.erl
evanmcc: added

src/miner.erl
evanmcc: yes, this should cause nothing to be put into the proto for a bytes field, but we should make sure that a mixed cluster can make progress.

test/miner_blockchain_SUITE.erl
```diff
@@ -26,7 +26,9 @@ all() -> [
           group_change_test,
           master_key_test,
           version_change_test,
-          election_v3_test
+          election_v3_test,
+          snapshot_test,
+          high_snapshot_test
```
Vagabond, line 31: Do we need a test where the snapshot is fixing a corrupted ledger?

src/miner.erl
Vagabond: This has been confirmed

PR #391, opened 2020-05-11
Don't challenge without GPS lock close to assert

evanmcc:
This is a WIP version to get some feedback on the approach here.

The idea is that we skim GPS info off of our status packet and store it in the `miner_lora` server.  Then we check if it's close enough to our asserted position to see if we're allowed to make any requests.

Main questions here: 
1. Is 75 meters too lenient? Too strict?
2. Some other GPS using PRs we do some additional position adjustment, do we need to do those here?

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -94,17 +94,17 @@ position() ->
 location_ok() ->
     %% the below code is tested and working.  we're going to roll
     %% out the metadata update so app users can see their status
-    case position() of
-        {error, _Error} ->
-            lager:debug("pos err ~p", [_Error]),
-            false;
-        {ok, _} ->
-            true;
-        %% fix but too far from assert
-        {ok, _, _} ->
-            false
-    end.
-    %% true.
+    %% case position() of
+    %%     {error, _Error} ->
+    %%         lager:debug("pos err ~p", [_Error]),
+    %%         false;
+    %%     {ok, _} ->
+    %%         true;
+    %%     %% fix but too far from assert
+    %%     {ok, _, _} ->
+    %%         false
+    %% end.
+    true.
 
 %% ------------------------------------------------------------------
 %% gen_server Function Definitions
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -7,7 +7,8 @@
     handle_response/1,
     send/1,
     send_poc/5,
-    port/0
+    port/0,
+    position/0
 ]).
 
 -export([
@@ -40,7 +41,8 @@
     packet_timers = #{}, %% keyed by token
     sig_fun,
     pubkey_bin,
-    mirror_socket
+    mirror_socket,
+    latlong
 }).
 
 -type state() :: #state{}.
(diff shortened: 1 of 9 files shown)


fvasquez:
> 1. Is 75 meters too lenient? Too strict?

I think 75 meters may be too lenient. On average, smartphone GPS location accuracy is 30 meters in dense urban areas.

evanmcc:
I guess I'd rather start too lenient and crank it down if too many people are slipping through.  With 30m accuracy, 75m might not be lenient enough.

evanmcc:
commit 83fe6b9 will require merging https://github.com/helium/blockchain-core/pull/445 first to work

commit bc12b29 should be reverted for testing and to make the feature active

fasuloj:
I've left gps trackers in my parked vehicle dashboard and it has veered off a quarter mile in any direction on cargo. Cargo shows it as if if moved around the area even though it is in a stationary position.. I would expect the same drift from my hotspots GPS position.

evanmcc:
Hi @fasuloj!

Rest assured that there will be a long data gathering and analysis period before we deploy this change widely.  The PR as currently written implements the feature in a kind of data gathering mode.  It adds anonymized informatoin to the gossip metadata, so we can figure out how many people would be affected by this.  There will be more official communications in the coming weeks about how the deployment is going to go.

src/miner_lora.erl
```diff
@@ -131,6 +140,36 @@ handle_call({send, Payload, When, Freq, DataRate, Power, IPol}, From, #state{soc
...
+                undefined ->
+                    {error, not_asserted};
+                H3Loc ->
+                    case vincenty:distance(h3:to_geo(H3Loc), State#state.latlong) of
+                        {error, _E} ->
+                            lager:debug("fix error! ~p", [_E]),
+                            {error, bad_calculation};
+                        {ok, Distance} when Distance > ?MAX_WANDER_DIST ->
+                            lager:debug("fix too far! ~p", [Distance]),
+                            {ok, bad_assert, State#state.latlong};
```
vihu, line 161: This return breaks the type spec contract

src/poc/miner_poc_statem.erl
```diff
@@ -623,28 +622,40 @@ allow_request(BlockHash, #data{blockchain=Blockchain,
...
+                    true;
+                LastChallenge ->
+                    case (Height - LastChallenge) > POCInterval of
+                        true -> 1 == rand:uniform(trunc(POCInterval / 4));
+                        false -> false
                     end
-            end
+            end,
+        LocationOK = true,
+        %% the below code is tested and working.  we're going to roll
```
Vagabond, line 641: You should probably also not report challenges or witnesses if you have this issue as well.

evanmcc:
6763e04 is the new commit to revert to activate this feature.

src/poc/miner_poc_statem.erl
evanmcc: should be remedied by a167a2d

PR #402, opened 2020-05-19
Docker tweaks

lthiery:
The main changes here are that the miner is run in foreground, allowing `docker run` to return when `-d` option is used.

In addition, `/opt/miner/bin` is added to PATH so that interrogating miner is simpler, ie:

`docker exec happy_lamport miner info height`

rather than 

`docker exec happy_lamport /opt/miner/bin/miner info height`

Also, a teensy bit of documentation is added about how to build a test Docker image locally.

Fuller documentation on the docker will be available on developer.helium.com


mrallen1:
:+1: for the record

PR #405, opened 2020-05-22
Add support for 868 channel mapping

andymck:
related PRs:
https://github.com/helium/routerv3/pull/81
https://github.com/helium/blockchain-core/pull/473
https://github.com/helium/proto/pull/34

This adds region/location checks to miner_lora.erl.  The server will initialise an ets table with a bunch of country specific data ( read from a newly added csv file ).  The csv file provides frequency data on a per country basis.  

After init the server will:

1.  Check if location is asserted
2.  If false then block transmitting and recheck for location assert after 30 seconds 
3.  If true then pull geolocation data from `https://api.helium.io/v1/hotspots/<Addr>`
4.  Using the country code pulled from step 3, index into ets table and retrieve regionalised frequency data
5.  Store the relevant frequency data in server state and proceed to enable transmitting
6.  Use the region specific frequency in all subsequent transmitting
7.  All packets routed to state channels will now include the region which in turn will be sent to router and from there the relevant frequency/channel utilised

To be compatible with miner pro ( which apparently will not assert location ), the region checks can be overridden via miner env var `override_reg_domain_check`, set this too true to enable override.  In doing so the following additional params also need to be specified:

default_reg_region:  <<"US902-928">> | <<"EU863-870">> etc
default_reg_geo_zone: <<"zone1">> | <<"zone2">> | <<"zone3">>
default_reg_freq_list: [868.10, 868.30, 868.50] etc

When override is enabled, the miner will be able to transmit from init.


config/sys.config
```diff
@@ -75,6 +75,20 @@
...
+   {frequency_data, #{<<"US915">> => [903.9, 904.1, 904.3, 904.5, 904.7, 904.9, 905.1, 905.3],
+                      <<"EU868">> => [867.1, 867.3, 867.5, 867.7, 867.9, 868.1, 868.3, 868.5, 868.8],
+                      <<"EU433">> => [433.175, 433.375, 433.575],
+                      <<"CN470">> => [486.3, 486.5, 486.7, 486.9, 487.1, 487.3, 487.5, 487.7 ],
+                      <<"CN779">> => [779.5, 779.7, 779.9],
+                      <<"AU915">> => [916.8, 917.0, 917.2, 917.4, 917.5, 917.6, 917.8, 918.0, 918.2],
+                      <<"AS923">> => [923.2, 923.4, 923.6, 923.8, 924.0, 924.2, 924.4, 924.5, 924.6, 924.8],
+                      <<"KR920">> => [922.1, 922.3, 922.5, 922.7, 922.9, 923.1, 923.3],
+                      <<"IN865">> => [865.0625, 865.4025, 865.985],
+                      <<"zone1">> => [867.1, 867.3, 867.5, 867.7, 867.9, 868.1, 868.3, 868.5, 868.8],
```
Vagabond, line 89: I'm confused as to the relationship between eg. US915 and zone2

Vagabond:
So, there's a bunch of disagreement on the region names and their type. The enums in the proto are atoms, so we should just use those atoms everywhere and kill the binary form. I fixed some of this in core, but it's in plenty of other places as well..

Vagabond:
I fixed some more of it in core and in router, but it's still not perfect.

The overall approach looks promising though.

config/sys.config
andymck: Some countries do not have a region frequency specified, in these cases we default to a zone and for each zone there is a default frequency.

Vagabond: I'm not sure if we should fall back to 'zone' as I don't know how legal that is?

andymck: If a spot ends up in a country with no specified plan how should we deal with it then ?  Its either transmit on some specified plan or block forever

config/sys.config
```diff
@@ -76,6 +76,20 @@
                     {127,0,0,1}, 31341} },
    {default_routers, ["/p2p/112qB3YaH5bZkCnKA5uRH7tBtGNv2Y5B4smv1jsmvGUzgKT71QpE", "/p2p/1124CJ9yJaHq4D6ugyPCDnSBzQik61C1BqD9VMh1vsUmjwt16HNB"]},
    {mark_mods, [miner_hbbft_handler]},
-   {stabilization_period_start, 187000}
+   {stabilization_period_start, 187000},
+   {reg_domains_file, "countries_reg_domains.csv"},
+   {frequency_data, #{'US915' => [903.9, 904.1, 904.3, 904.5, 904.7, 904.9, 905.1, 905.3],
+                      'EU868' => [867.1, 867.3, 867.5, 867.7, 867.9, 868.1, 868.3, 868.5, 868.8],
```
Vagabond, line 82: Aren't there 9 channels here?

andymck: 868.8 prob should not be there, its listed as  FSK on things network

config/sys.config
andymck: it was agreed that we should block

src/cli/miner_cli_info.erl
```diff
@@ -199,6 +201,30 @@ info_p2p_status(["info", "p2p_status"], [], []) ->
...
+    [["info", "region"],
+     ["info region \n\n",
+      "  Get the frequency region this miner is operating in \n\n"
+      "  If undefined, the region has not yet been confirmed or cannot be determined\n"
+      "  Until a region is confirmed, the miner will block tx\n"
+     ]
+    ].
+
+info_region(["info", "region"], [], []) ->
+    {ok, Region} = miner_lora:region(),
```
Vagabond, line 223: You can do like

```erlang
case miner_lora:region() of
    {ok, undefined} ->
        [clique_status:alert("undefined")];
    {ok, Region} ->
         [clique_status:text(atom_to_list(Region))]
    end;
```

JayKickliter:
Is this PR ready to merge, or is it blocked?

andymck:
@JayKickliter it will be merged in the next hour.   Waiting on currently running tests to complete so as to verify latest rebase

andymck:
@JayKickliter merged

JayKickliter:
thanks!

PR #412, opened 2020-06-01
Step-by-step miner building guide

lthiery:
This was good before we had Docker, but now that we have Docker, I want to make sure people are only building it if they really really want to. Bouncing them from the devdocs to git will be a start, IMO.

We could also remove this entirely, but it might be nice to have something more verbose than what is currently available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [![Build Status](https://badge.buildkite.com/a2ced4f1160fa02aa8b735e7edb80f8ef787a299963ff88942.svg?branch=master)](https://buildkite.com/helium/miner)
 
-miner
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,3 +28,51 @@ Build a miner-test image locally:
...
+
+Clone the git repository:
+
+```bash
+git clone https://github.com/helium/miner.git
+```
+
+### Install Erlang
+
+Miner has been tested against Erlang OTP 21.3.3.
```
Vagabond, line 55: 22 is better, really

lthiery: Ok done

PR #429, opened 2020-06-17
adjust cache limits

evanmcc:
support for https://github.com/helium/blockchain-core/pull/506


config/sys.config
```diff
@@ -36,7 +36,9 @@
    {outbound_gossip_connections, 2},
    {peerbook_allow_rfc1918, false},
    {metadata_fun, fun miner_util:metadata_fun/0},
-   {relay_limit, 50}
+   {relay_limit, 50},
+   {gw_cache_retention_limit, 30},
```
vihu, line 40: Is this what we want to set on prod?

PR #430, opened 2020-06-17
on restore, we can't use the current ledger

evanmcc:
when we restore an older election, we need to use the ledger at that height, otherwise we might get the wrong group and think erroneously that we're not in the election group.

### DIFF
--- a/src/miner_consensus_mgr.erl
+++ b/src/miner_consensus_mgr.erl
@@ -715,7 +715,7 @@ code_change(_OldVsn, State, _Extra) ->
 
 initiate_election(Hash, Height, #state{delay = Delay} = State) ->
     Chain = blockchain_worker:blockchain(),
-    Ledger = blockchain:ledger(Chain),
+    {ok, Ledger} = blockchain:ledger_at(Height + Delay, Chain),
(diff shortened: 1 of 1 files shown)


src/miner_consensus_mgr.erl
```diff
@@ -715,7 +715,7 @@ code_change(_OldVsn, State, _Extra) ->
 
 initiate_election(Hash, Height, #state{delay = Delay} = State) ->
     Chain = blockchain_worker:blockchain(),
-    Ledger = blockchain:ledger(Chain),
+    {ok, Ledger} = blockchain:ledger_at(Height + Delay, Chain),
```
vihu, line 718: Do we want to catch or case clause any of these calls to `ledger_at`?

evanmcc: I mean, if that fails, does anything that we're doing make any sense?  I guess we could make it so that we don't crash, but honestly I'm not sure I see the value.

vihu: Yeah, just wanted to ask. Maybe it's better to not crash? We're definitely not gonna progress if `ledger_at` fails though

PR #436, opened 2020-06-22
Easily Override Regions in Docker

lthiery:
It's hard for people to route their first packets with the new region configuration scheme.

This makes it easy to configure the override when starting the docker container.

Long-term, I wonder if we can come up with a better solution for: "how can I forward packets with a miner without it being added/asserted yet"? Of course, I would not expect to earn if I don't add, but I should be able to forward packets for free. In addition, I would expect that I could earn but not participate in POC if I am added but with no location asserted.

I wonder if the miner could simply read in an environmental variable that would allow this kind of override to happen?

### DIFF
--- a/.buildkite/scripts/Dockerfile-amd64
+++ b/.buildkite/scripts/Dockerfile-amd64
@@ -20,6 +20,20 @@ RUN tar -zxvf _build/prod/rel/*/*.tar.gz -C /opt/rel
 RUN sed -i 's/^\s*{key,/%{key,/' /opt/rel/releases/0.1.0/sys.config
 RUN sed -i 's/{use_ebus, true},/{use_ebus, false},/' /opt/rel/releases/0.1.0/sys.config
 RUN sed -i 's/{radio_device, { {127,0,0,1}, 1680,/{radio_device, { {0,0,0,0}, 1680,/' /opt/rel/releases/0.1.0/sys.config
+
+# Place start script 
+COPY .buildkite/scripts/start.sh /opt/rel/start.sh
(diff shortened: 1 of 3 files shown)


andymck:
@lthiery its certainly doable to rely on env vars here to configure the region data.  And it can prob be done via the use of a single env var rather than the 4 current vars.  This is for miner pro only though, right?

lthiery:
Yes, this use-case is for miner pro principally. People using our Hotspot can add easily so its less of an issue. There is maybe an edge-case for people who choose to not assert location, but I don't think that's come up yet.

If we went the single env variable route, it could be slightly less janky than this PR. I don't like that the Dockerfile derives US/EU `sys.configs` from a base `sys.config` as that seems easy to mess up with upstream changes. Furthermore, this PR copies the appropriate one on Docker boot, so if a user wanted to customize the config file further, they have to realise that `sys.config` is not the one to customize but instead the one that is going to be copied over during the next boot.

Another idea (rather than an environmental variable) which might be generally beyond this use-case is to check the "base_dir" for "local.config" where I could have specific field overrides relative to  "sys.config". For example, I could preserve my validation_width settings that way or the amount of peers I want.

lthiery:
I've moved the "alternate solutions" discussion to a Clubhouse ticket: https://app.clubhouse.io/hlm/story/8039

I would suggest landing this PR and reverting when a cleaner solution is available.

lthiery:
Seems like @Vagabond is working on the cleaner solution so I'm going to withdraw this PR

PR #437, opened 2020-06-25
expose PULL_RESP frames to info logs

lthiery:
It's not possible to see downlinks with the current log configuration. This mirrors the `info` level of PUSH_DATA (uplinks) with PULL_RESP frames output at the `info` level.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -222,9 +222,7 @@ handle_call({send, Payload, When, ChannelSelectorFun, DataRate, Power, IPol}, Fr
             %% the fun is set by the sender and is used to deterministically route data via channels
             LocalFreq = ChannelSelectorFun(Freqs),
             Token = mk_token(Timers),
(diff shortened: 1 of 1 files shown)


src/miner_lora.erl
```diff
@@ -247,6 +247,9 @@ handle_call({send, Payload, When, ChannelSelectorFun, DataRate, Power, IPol}, Fr
                              <<"data">> => base64:encode(Payload)
                             }
                         }),
+
+            lager:info("PULL_RESP ~p to ~p:~p", [jsx:decode(BinJSX), IP, Port]),
```
andymck, line 251: as BinJSX is encoded in the previous line, it would be good to declare the unencoded version as a new var and use that in the lager statement...save having to decode a payload we  already have in unencoded state

lthiery: Ah good point. Just pushed a commit to remedy

lthiery:
anything holding this puppy up? @andymck

PR #439, opened 2020-06-25
Factor out update_gateway_record into fn

lthiery:
I was trying to reason about [the ticket I made](https://app.clubhouse.io/hlm/story/8047/miner-should-remember-address-of-new-radio-clients) and I found an interesting snippet that looked worthy of factoring out due to repetition.

What's even more interesting is that this code seems to suggest that my desired behavior in my ticket should happen... I'm going to try to test out the issue more rigorously, but this mini-PR seems OK?


src/miner_lora.erl
```diff
@@ -375,6 +375,21 @@ select_gateway(Gateways) ->
             {ok, erlang:element(2, erlang:hd(maps:to_list(Gateways)))}
     end.
 
+%% Updates (IP, Port) for a given MAC, increments received packets from that MAC
+%% and creates gateway if necessary
+-spec update_gateway_record(map(), binary(), inet:ip_address(), inet:port_number()) -> gateway().
+update_gateway_record(
+    Gateways,
+    <<MAC:64/integer>>,
```
andymck, line 383: Mac here should just be a standard param, no need for bit syntax here.  Bit syntax was already used to derive the value from the packet data in the caller.

evanmcc: agree with this, as it could be semantics-changing.

PR #445, opened 2020-06-30
remove ledger check to see if GW exists, go straight to API when retrieving geo info

andymck:


evanmcc:
this looks fine, but I guess I'd rather have it work something like:

1. check ledger: if asserted, use the asserted location without an external call
2. fallback to external call if not asserted

I'm not sure why the original code doesn't seem to use the asserted location.  if it's not possible to do this all locally and we rely on the API for some of this information, ignore this.

andymck:
so the asserted location is a h3 ref which does not give us what we need, ie we need to know which country the hotspot is located in.  The only means we have to get the country (without using an external/non helium API ) is from the existing API data.

Vagabond:
> OK. extremely not wild about being reliant on our API. I think that a medium term goal should be to do this entirely on-board when we have an asserted location, but this is an improvement on the status quo.

Yeah, this was always a poor compromise, and we should revisit.

andymck:
yip, relying on the API was a V1 solution which got us over the line...

PR #446, opened 2020-07-01
add var 31 and backfill 14 and 15

evanmcc:
some audit paperwork

### DIFF
--- a/audit/var-31.md
+++ b/audit/var-31.md
@@ -1,8 +1,9 @@
-# Chain Variable Transaction 30
+# Chain Variable Transaction 31
 
 ## Changes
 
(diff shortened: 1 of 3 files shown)


audit/var-31.md
```diff
@@ -0,0 +1,37 @@
+# Chain Variable Transaction 30
+
+## Changes
+
+This txn enables support for transaction fees for most blockchain transactions. Additionally we've retired one price oracle key and replaced it with a new one.
```
Vagabond, line 5: I don't think this is right?

audit/var-31.md
```diff
@@ -0,0 +1,37 @@
+# Chain Variable Transaction 30
```
Vagabond, line 1:
```suggestion
# Chain Variable Transaction 31
```

PR #450, opened 2020-07-08
rework supervisors

evanmcc:
make it so when the blockchain or the consensus manager is sad, the whole node just bails immediately.

### DIFF
--- a/src/miner_sup.erl
+++ b/src/miner_sup.erl
@@ -16,7 +16,7 @@
     id => I,
     start => {I, start_link, Args},
     restart => permanent,
-    shutdown => 5000,
+    shutdown => infinity,
     type => supervisor,
     modules => [I]
 }).
--- a/src/miner_sup.erl
+++ b/src/miner_sup.erl
@@ -21,15 +21,6 @@
     modules => [I]
 }).
 
--define(WORKER(I, Args), #{
-    id => I,
-    start => {I, start_link, Args},
-    restart => permanent,
-    shutdown => 5000,
-    type => worker,
-    modules => [I]
-}).
-
 %% ------------------------------------------------------------------
 %% API functions
 %% ------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/miner_sup.erl
```diff
@@ -41,118 +32,14 @@ start_link() ->
...
-                RegionOverRide = check_for_region_override(),
-                OnionOpts = #{
-                    radio_udp_bind_ip => RadioBindIP,
-                    radio_udp_bind_port => RadioBindPort,
-                    radio_udp_send_ip => RadioSendIP,
-                    radio_udp_send_port => RadioSendPort,
-                    ecdh_fun => ECDHFun,
-                    sig_fun => SigFun,
-                    region_override => RegionOverRide
+                 strategy => rest_for_one,
```
andymck, line 35: so if a process under the critical sup goes down, the miner stops.  Seems strange to have that as the desired outcome.  why not allow the entire tree to restart rather than rely on external intervention ( canary or human ) ?

evanmcc: that's the current behavior, and for whatever reason it doesn't seem to work particularly well.  this is definitely a bandaid.

andymck: well not exactly...i think i commented on the wrong line originally so i maybe didnt convey things well...

The old top level sup `miner_sup` had a restart strategy of 

intensity => 2,
period => 5

The new has

intensity => 0,
period => 1

And as the miner_critical_sup has the same new restart strategy... if any of those processes under the critical sup, such as blockchain_worker, go down then both the critical sup and the miner sup restart strategies are breached taking down the miner.

In the old way for the same scenario the restart strategy would at least attempt to restart the miner tree 2 times.  Now there are no automatic restarts off the miner tree.  

Is that the desired new behaviour ?

andymck: oh i have just realised the sups here are using a rest for one and were doing so in the original which meant the entire tree was unlikely to ever be restarted unless there was a breach of the restart policy.

if you changed the miner_sup to a one_for_all and a restart strategy of 5, 2 it would ensure the entire miner tree is restarted if something in the critical_sup goes down.

I dont know about the exact issue/scenario are you trying to address here but it could be the rest_for_one is resulting in some non clean behaviour during one of more process crashes..

andymck: i tested the above approach locally and it works fine ( assuming you want the entire miner_sup tree restarted if something in the critical_sup goes down )...
Works fine except for one major caveat, during the restart the swarm sup fails to start and reports an already started error...so something is not right with how these processes are being supervised.

If you put the miner_sup intensity way up ( say 200 ), it eventually takes care of itself as the swarm_sup eventually dies and permits the restart of the miner tree

I am pretty sure i fixed this in one of my old unmerged prs addresses shutdown behaviour....i can have a look at it again later.

This lack of clean shutdown may be part of the root issue that was the driver behind this PR in the first place ?

andymck: the swarm is started from blockchain_swarm and killed there also as part of blockchain_swarm's terminate
but as blockchain_swarm is not trapping exits the terminate is not called during a shutdown.
Adding a trap exit here gives a clean shutdown and enables the full miner_tree to be restarted cleanly in the event of one of the critical_sup's processes going down and avoids the already_started error

andymck: i should prob revisit those clean shutdown PRs and get them up to date and we should try to get them merged

andymck: To summarise my comments I am suggesting the following additional changes:

1. Make miner_sup a one_for_all type with a restart strategy of 5, 2
2. Update blockchain_swarm.erl to trap exits to ensure its terminate is called during a shutdown, restart etc which will ensure the swarm sup is stopped cleanly

Together this will ensure that should a process under the criticial_sup go down the entire miner tree will be restarted cleanly

PR #454, opened 2020-07-17
Add downlink throttling

JayKickliter:
This PR adds region-aware enforcement of transmission restrictions.

## Still needed

- [ ] EU868 is definitely not correct. I'm still a little fuzzy on ETSI duty-cycle rules, and some subbands allow more/less duty-cycle than others
- [ ] US915 should take into account TX bandwidth. Right now the throttle considers the 400ms dwell time over 20s, but that can be reduced to 400ms over 10s for transmissions at bandwidth >=250 kHz

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -49,7 +49,8 @@
     latlong,
     reg_domain_confirmed = false :: boolean(),
     reg_region :: atom(),
-    reg_freq_list :: [float()]
+    reg_freq_list :: [float()],
+    reg_throttle :: miner_lora_throttle:handle()
 }).
 
 -record(country, {
@@ -204,7 +205,8 @@ init(Args) ->
                 pubkey_bin = blockchain_swarm:pubkey_bin(),
                 reg_domain_confirmed = RegDomainConfirmed,
                 reg_region = DefaultRegRegion,
-                reg_freq_list = DefaultRegFreqList}}.
+                reg_freq_list = DefaultRegFreqList,
+                reg_throttle = miner_lora_throttle:new(DefaultRegRegion)}}.
 
 handle_call({send, _Payload, _When, _ChannelSelectorFun, _DataRate, _Power, _IPol}, _From,
                 #state{reg_domain_confirmed = false}=State) ->
@@ -214,7 +216,8 @@ handle_call({send, Payload, When, ChannelSelectorFun, DataRate, Power, IPol}, Fr
                 #state{ socket=Socket,
                         gateways=Gateways,
                         packet_timers=Timers,
-                        reg_freq_list = Freqs}=State) ->
+                        reg_freq_list = Freqs,
+                        reg_throttle = Throttle}=State) ->
     case select_gateway(Gateways) of
         {error, _}=Error ->
             {reply, Error, State};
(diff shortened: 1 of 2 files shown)


src/miner_lora_throttle.erl
```diff
@@ -0,0 +1,241 @@
...
+%% limited regions (EU868).
+%%
+%% In order to calculate duty cycle, we track every single
+%% transmission 'now' and the previous DUTY_CYCLE_PERIOD_MS of
+%% time. Note that 'now' is always changing and that transmissions
+%% older than DUTY_CYCLE_PERIOD_MS are only untracked when updating
+%% state or calculating duty-cycle.
+-define(DUTY_CYCLE_PERIOD_MS, 3600000).
+
+%% Maximum time on air for dell-time limited regions (US915).
```
mrallen1, line 46: dwell typo

src/miner_lora_throttle.erl
```diff
@@ -0,0 +1,241 @@
...
+%% Scenario 4: 100 % of packet transmission after CutoffTime.
+dwell_time([P | T], CutoffTime, Frequency, Acc) ->
+    dwell_time(T, CutoffTime, Frequency, Acc + P#sent_packet.time_on_air);
+dwell_time([], _CutoffTime, _Frequency, Acc) ->
+    Acc.
+
+%% @doc Returns total time on air for packet sent with given
+%% parameters.
+%%
+%% See Semtech Appnote AN1200.13, "LoRa Modem Designer's Guide"
```
mrallen1, line 169: It'd be nice if you included a short summary of the note below this comment

src/miner_lora_throttle.erl
```diff
@@ -0,0 +1,241 @@
...
+-spec time_on_air(
+    Bandwidth :: number(),
+    SpreadingFactor :: number(),
+    CodeRate :: integer(),
+    PreambleSymbols :: integer(),
+    ExplicitHeader :: boolean(),
+    PayloadLen :: integer()
+) ->
+    Milliseconds :: float().
```
mrallen1, line 178: I thought millis would be integers?

JayKickliter: The math itself inherently returns floating point values. I could take the ceiling if it made sense. The only reason why this module uses milliseconds everywhere is that is the unit all the regulatory bodies use.

src/miner_lora_throttle.erl
JayKickliter: I'd be happy to, but I'm not sure what what there is to summarize. The function body is cut & paste from the app note and erlangified.

src/miner_lora_throttle.erl
JayKickliter: thanks. Emacs is supposed to catch these, but dell is word.

src/miner_lora_throttle.erl
Vagabond: Maybe just include a link?

src/miner_lora_throttle_tests.erl
```diff
@@ -0,0 +1,125 @@
+-module(miner_lora_throttle_tests).
```
Vagabond, line 1: I think this module should be in test/ not src/

src/miner_lora.erl
```diff
@@ -414,66 +449,67 @@ handle_udp_packet(<<?PROTOCOL_2:8/integer-unsigned,
...
-            State#state{packet_timers=maps:remove(Token, Timers)};
-        {ok, {send, Ref, From}} ->
+            State#state{packet_timers=maps:remove(Token, Timers),
+                        reg_throttle=miner_lora_throttle:track_sent(Throttle, SentAt, LocalFreq, TimeOnAir)};
+        {ok, {send, Ref, From, SentAt, LocalFreq, TimeOnAir}} ->
             %% likely some kind of error here
             _ = erlang:cancel_timer(Ref),
-            Reply = case kvc:path([<<"txpk_ack">>, <<"error">>], jsx:decode(MaybeJSON)) of
+            {Reply, NewThrottle} = case kvc:path([<<"txpk_ack">>, <<"error">>], jsx:decode(MaybeJSON)) of
                 <<"NONE">> ->
```
Vagabond, line 464: Does the concentrator return this immediately if the packet is queued, or when the packet actually gets sent?

src/miner_lora_throttle.erl
```diff
@@ -0,0 +1,224 @@
...
+%%
+-spec can_send(
+    Handle :: handle(),
+    AtTime :: number(),
+    Frequency :: integer(),
+    TimeOnAir :: number()
+) ->
+    boolean().
+can_send(_Handle, _AtTime, _Frequency, TimeOnAir) when TimeOnAir > ?MAX_TIME_ON_AIR_MS ->
+    %% TODO: check that all regions have do in fact have the same
```
Vagabond, line 122: Is there a way to resolve this?

src/miner_lora.erl
JayKickliter: as soon as it's [queued](https://github.com/helium/packet_forwarder/blob/master/lora_pkt_fwd/src/lora_pkt_fwd.c#L2539-L2549)

src/miner_lora_throttle.erl
JayKickliter: The rules are a bit nebulous, and it seems like everyone interprets them differently.  From what I can tell, it is resolved, and this will not allow for too-long-on-air. I do not _think_ it will artificially restrict time on air to what's allowed, but not 100% certain.

Vagabond:
Ok, I think we're clear to merge this, we just have to figure out if we merge it before/after the rx2 work.

Vagabond (approved):

PR #459, opened 2020-07-22
[DNM] Test state channels v2 with updated offer definition

vihu:


vihu:
No longer necessary, closing.

PR #487, opened 2020-08-15
Bump for new core and sc client fixes

vihu:


rebar.config
```diff
@@ -10,7 +10,7 @@
 
 {deps, [
         {blockchain, {git, "https://github.com/helium/blockchain-core.git",
-                      {branch, "master"}}},
+                      {branch, "rg/more-sc-client-bug-fixes"}}},
```
vihu, line 13: Switch back to master when it lands

PR #491, opened 2020-08-24
add per-type timeouts to txns

evanmcc:
i think that a lot of the time issues we're seeing are coming from certain receipts becoming too large to validate in a timely fashion, so this puts much tighter restrictions on them.

### DIFF
--- a/src/miner_hbbft_sidecar.erl
+++ b/src/miner_hbbft_sidecar.erl
@@ -249,7 +249,7 @@ code_change(_OldVsn, State, _Extra) ->
 filter_txn_buffer(Buf, Chain) ->
     lists:filter(fun(BinTxn) ->
                          Txn = blockchain_txn:deserialize(BinTxn),
-                         IsSlow = error /= maps:find(blockchain_txn:type(Txn), ?SlowTxns),
+                         IsSlow = maps:is_key(blockchain_txn:type(Txn), ?SlowTxns),
                          case IsSlow orelse blockchain_txn:is_valid(Txn, Chain) == ok of
                              true ->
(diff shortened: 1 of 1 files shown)


src/miner_hbbft_sidecar.erl
```diff
@@ -284,15 +284,16 @@ maybe_start_validation(#state{queue = Queue, chain = Chain,
         [] ->
             State;
         [{From, Txn} | Queue1] ->
-            {Attempt, V} = start_validation(Txn, From, Chain),
+            Type = blockchain_txn:type(Txn),
+            Timeout = maps:get(Type, ?SlowTxns, 10000),
```
Vagabond, line 288:
```suggestion
            Timeout = maps:get(Type, ?SlowTxns, application:get_env(miner, txn_validation_budget_ms, 10000)),
```

src/miner_hbbft_sidecar.erl
```diff
@@ -249,7 +249,7 @@ code_change(_OldVsn, State, _Extra) ->
 filter_txn_buffer(Buf, Chain) ->
     lists:filter(fun(BinTxn) ->
                          Txn = blockchain_txn:deserialize(BinTxn),
-                         IsSlow = lists:member(blockchain_txn:type(Txn), ?SlowTxns),
+                         IsSlow = error /= maps:find(blockchain_txn:type(Txn), ?SlowTxns),
```
Vagabond, line 252:
```suggestion
                         IsSlow = maps:is_key(blockchain_txn:type(Txn), ?SlowTxns),
```

PR #497, opened 2020-08-30
Read onboarding_key from file on Blackspots

fvasquez:
The Blackspots that RAK is manufacturing for the Helium Network will come with a version 4 UUID written to an `onboarding_key` file on the boot partition of the microSD card.

```
# cat /etc/hardware_version 
blackspot
# cat /mnt/uboot/onboarding_key 
e8d57c09f96c41229ff34f8889504a05
# miner print_keys
{pubkey,"11k4Gn6Qhs6JDet1w1fKxUe647zw4KRiZCYhNBtqYCL16BNmjRo"}.
{onboarding_key,"e8d57c09f96c41229ff34f8889504a05"}.
{animal_name,"decent-tortilla-bee"}.
```

Notice that the `onboarding_key` displayed for a Blackspot is no longer `undefined` when a `/mnt/uboot/onboarding_key` file exists.

### DIFF
--- a/src/miner_keys.erl
+++ b/src/miner_keys.erl
@@ -20,6 +20,8 @@
 -spec keys({file, BaseDir::string()} |
            {ecc, proplists:proplist()}) -> key_info().
 keys({file, BaseDir}) ->
+    BootDir = "/mnt/uboot",
+    OnboardingKey = filename:join([BootDir, "onboarding_key"]),
     SwarmKey = filename:join([BaseDir, "miner", "swarm_key"]),
     ok = filelib:ensure_dir(SwarmKey),
     case libp2p_crypto:load_keys(SwarmKey) of
@@ -28,7 +30,14 @@ keys({file, BaseDir}) ->
                key_slot => undefined,
                ecdh_fun => libp2p_crypto:mk_ecdh_fun(PrivKey0),
                sig_fun => libp2p_crypto:mk_sig_fun(PrivKey0),
-               onboarding_key => undefined
+               onboarding_key => case filelib:ensure_dir(OnboardingKey) of
+                                     ok ->
+                                         case file:read_file(OnboardingKey) of
+                                             {ok, Bin} -> Bin;
+                                             {error, _Reason} -> undefined
+                                         end;
+                                     {error, _Reason} -> undefined
+                                 end
              };
         {error, enoent} ->
             KeyMap = #{secret := PrivKey0, public := PubKey} = libp2p_crypto:generate_keys(ecc_compact),
@@ -37,7 +46,14 @@ keys({file, BaseDir}) ->
                key_slot => undefined,
                ecdh_fun => libp2p_crypto:mk_ecdh_fun(PrivKey0),
                sig_fun => libp2p_crypto:mk_sig_fun(PrivKey0),
-               onboarding_key => undefined
+               onboarding_key => case filelib:ensure_dir(OnboardingKey) of
+                                     ok ->
+                                         case file:read_file(OnboardingKey) of
+                                             {ok, Bin} -> Bin;
+                                             {error, _Reason} -> undefined
+                                         end;
+                                     {error, _Reason} -> undefined
+                                 end
              }
     end;
 keys({ecc, Props}) when is_list(Props) ->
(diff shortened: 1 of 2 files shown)


src/miner_keys.erl
```diff
@@ -20,6 +20,8 @@
 -spec keys({file, BaseDir::string()} |
            {ecc, proplists:proplist()}) -> key_info().
 keys({file, BaseDir}) ->
+    BootDir = "/mnt/uboot",
+    OnboardingKey = filename:join([BootDir, "onboarding_key"]),
```
madninja, line 24: So two options:

1. Rename to OnboardingDir and get it from sys.config instead of hardcoding here
2. Mount the folder in BaseDir somewhere and have OnboardingKey just offset from BaseDir

src/miner_keys.erl
```diff
@@ -28,7 +30,14 @@ keys({file, BaseDir}) ->
                key_slot => undefined,
                ecdh_fun => libp2p_crypto:mk_ecdh_fun(PrivKey0),
                sig_fun => libp2p_crypto:mk_sig_fun(PrivKey0),
-               onboarding_key => undefined
+               onboarding_key => case filelib:ensure_dir(OnboardingKey) of
```
madninja, line 33: Why are you trying to create the directory. Check if it's there, don't ensure its creation .

In fact only use the file:read_file to load it or fail?

src/miner_keys.erl
```diff
@@ -37,7 +46,14 @@ keys({file, BaseDir}) ->
...
```
madninja, line 49: Same as above.. also DRY: Move the block of code into a get_onboarding_key function

src/miner_keys.erl
```diff
@@ -94,8 +110,18 @@ print_keys(_) ->
     MaybeB58 = fun(undefined) -> undefined;
                   (Key) -> libp2p_crypto:pubkey_to_b58(Key)
                end,
+    MaybeUUIDv4 = fun(undefined) -> undefined;
+                  (Key) ->
+                      %% only production blackspots read onboarding_key from a file
+                      case filelib:is_file("/mnt/uboot/onboarding_key") of
```
madninja, line 116: Don't use the file, but catch the b58 encoding error and do the binary_to_list option?

madninja: And is MabeB58 dead code now?

src/miner_keys.erl
fvasquez: I am going to go with option 1.

src/miner_keys.erl
fvasquez: I thought `filelib:ensure_dir` only checked for the existence of a directory. I am newb.

src/miner_keys.erl
fvasquez: Will do.

src/miner_keys.erl
fvasquez: No, I believe we still need it for Helium Hotspots that read the `onboarding_key` out of the ECC608.

fvasquez: > Don't use the file, but catch the b58 encoding error and do the binary_to_list option?

Exceptions scare me. I actually encountered that b58 encoding error while implementing this function. I prefer this way.

src/miner_keys.erl
```diff
@@ -10,6 +10,23 @@
                      }.
 -export_type([key_info/0]).
 
+get_onboarding_filename() ->
+    case application:get_env(blockchain, onboarding_dir) of
```
madninja, line 14: Use application:get_env/3 to take the default value?

src/miner_keys.erl
```diff
@@ -94,8 +111,19 @@ print_keys(_) ->
...
+                      OnboardingKey = get_onboarding_filename(),
+                      case filelib:is_file(OnboardingKey) of
+                          true ->
+                              string:trim(binary_to_list(Key));
+                          false ->
+                              libp2p_crypto:pubkey_to_b58(Key)
+                      end
+               end,
     Props = [{pubkey, MaybeB58(PubKey)},
-             {onboarding_key, MaybeB58(OnboardingKey)}
```
madninja, line 98: Is Maybe B58 no longer used?

src/miner_keys.erl
fvasquez: > Use application:get_env/3 to take the default value?

I thought of doing that but then the default value would just be `undefined` which is almost the same as what I have now.

src/miner_keys.erl
fvasquez: > Is Maybe B58 no longer used?

It's used in the preceding line to validate `pubkey`. If I used `MaybeUUIDv4` instead then there would be no guarantee that `pubkey` is b58.

madninja: whoops, you're right , missed that one

PR #508, opened 2020-09-14
Change poc_id to use blockchain_utils

madninja:
uses the newly exposed `blockchain_utils:poc_id/1` to log the poc identifier consistently


vihu (approved):
Just remember to switch core branch and update lockfile once the corresponding core branch lands :+1:

PR #509, opened 2020-09-14
Bump blessed snapshot to height 498961

fvasquez:
```
# miner snapshot list
Height 498961
Hash <<187,98,112,202,202,35,102,203,98,12,16,120,12,216,24,188,223,66,65,40,
       61,162,124,63,64,194,74,224,103,80,123,109>>
Have true
```


fvasquez:
Use a newer snapshot.

PR #510, opened 2020-09-15
Bump blessed snapshot to height 499681

fvasquez:
```
# miner snapshot list
Height 499681
Hash <<4,171,167,134,123,207,42,239,46,240,121,72,154,56,75,108,198,45,134,40,
       252,107,219,206,219,31,38,163,9,244,31,179>>
Have true
```


vihu (approved):
lgtm

PR #516, opened 2020-09-17
Store request block hash in poc receipt

Vagabond:
Depends on helium/blockchain-core#622


test/miner_poc_SUITE.erl
```diff
@@ -70,7 +70,7 @@ all() ->
      poc_dist_v8_test,
      poc_dist_v8_partitioned_test,
      poc_dist_v8_partitioned_lying_test,
-     poc_dist_v9_test,
+     poc_dist_v10_test,
```
vihu, line 73: I seemed to have missed adding the other two v9 tests here, can you please add them for v10?

evanmcc: while you're in here, can we comment out the v7 tests?

src/poc/miner_poc_statem.erl
```diff
@@ -70,6 +70,7 @@
     responses = #{},
     receiving_timeout = ?RECEIVING_TIMEOUT :: non_neg_integer(),
     poc_hash  :: binary() | undefined,
+    request_block_hash  :: binary() | undefined,
```
evanmcc, line 73: default this to `<<>>`?  otherwise we'll get a lot of crashes on the transition.  might never see it, though.

evanmcc (approved):
other than the nits it looks good

src/poc/miner_poc_statem.erl
Vagabond: I'm not sure how that will help?

PR #519, opened 2020-09-22
Add a Docker Tag for latest-arch

lthiery:
Docker has a `latest` tag that is generally applied to repositories for the latest "GA" image. Community has been asking for this in some shape or form. I am thinking of PastaGringo in particular and @jamiew today exclaimed: "My kingdom for a ‘latest’ tag on quay".  Indeed, it would greatly simplify [scripts ](https://github.com/Wheaties466/helium_miner_scripts) meant to keep the service up to date.

In our case, our repository has Dockers for two architectures and so simply `latest` will not do. 

We could 
- split the repository, such that `https://quay.io/repository/team-helium/miner-amd64` and `https://quay.io/repository/team-helium/miner-arm64` exist
- append the architecture to the `latest` tag, such that tags `latest-amd64` and `latest-arm64` exist

The latter is less work, thus it is what I propose here. 

For testing, I created a presonal quay repository and tested the commands in this PR. My main concern was whether a previous tag could be overwritten without removing from the previous image as the documentation for [`docker tag`](https://docs.docker.com/engine/reference/commandline/tag/) was not explicit about this.

The first upload:

first_tag

The second upload:

second_tag

This appears to validate that it does overwrite the tag without much fuss.


fvasquez (approved):
I am glad to see the solution is that simple. Nice work.

PR #521, opened 2020-09-24
Dynamically select spreading factor for PoC packets

amirhaleem:
As Proof of Coverage onion packets are variable in size (based on the path length), we should dynamically select the correct spreading factor based on the packet size to ensure compliance with the LoRaWAN specification and any regional regulations.

### DIFF
--- a/audit/var-45.md
+++ b/audit/var-45.md
@@ -1,4 +1,4 @@
-# Chain Variable Transaction 45
+# Chain Variable Transaction 44
 
 ## Changes
 
--- a/audit/var-45.md
+++ b/audit/var-45.md
@@ -1,4 +1,4 @@
-# Chain Variable Transaction 44
+# Chain Variable Transaction 45
 
 ## Changes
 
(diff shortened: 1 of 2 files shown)


audit/var-45.md
```diff
@@ -1,4 +1,4 @@
-# Chain Variable Transaction 45
+# Chain Variable Transaction 44
```
vihu, line 1: Why was this changed? We're on var nonce 45...

src/poc/miner_onion_server.erl
```diff
@@ -391,3 +392,15 @@ try_decrypt(IV, OnionCompactKey, OnionKeyHash, Tag, CipherText, ECDHFun, Chain)
             {error, too_many_pocs}
     end.
 
+-spec spreading(integer()) -> string().
+spreading(Len) when Len < 12 ->
```
vihu, line 396: Just to confirm, on equality we want to go to the next matching fun?

audit/var-45.md
amirhaleem: Have no idea. Will remove this file

src/poc/miner_onion_server.erl
amirhaleem: Right. For example, if the packet size is 12 bytes, we should be using SF9BW125

vihu (approved):
looks good :+1:

PR #530, opened 2020-10-06
reduce the time period, be more aggressive about catchup

evanmcc:
While the current system draws neat curves, there's no inherent momentum to the system.  As such, we should move to a system that catches up at maximum speed until the final trim adjustments, which are made a minimum speed to prevent large time oscillations around the target time.

### DIFF
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -623,9 +623,9 @@ set_next_block_timer(State=#state{blockchain=Chain}) ->
     BlockTimeDeviation =
         case BlockTimeDeviation0 of
             N when N > 0 ->
-                catchup_time(N);
+                catchup_time(abs(N));
             N ->
-                -1 * catchup_time(N)
+                -1 * catchup_time(abs(N))
         end,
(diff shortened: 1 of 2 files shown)


src/miner.erl
```diff
@@ -652,3 +649,8 @@ process_bbas(N, BBAs) ->
                     <<>>
             end
     end.
+
+catchup_time(N) when N < 0.15 ->
```
vihu, line 653: Do we want equality here?

Vagabond: I don't think it really matters here. I *am* interested in what 0.15 actually means here though. I thought N was going to be in seconds.

evanmcc: N is a float, it's the result of `BlockTime  - (LastBlockTimestamp - StartBlockTime) / (Height - ActualStartHeight)`, which will be a float.  That said, there is a bug here: if the difference is 0, or very very small, we should return 0.

vihu: Is it possible that `Height - ActualStartHeight` can ever be 0 ?

Vagabond (approved):

Vagabond (approved):

PR #540, opened 2020-10-17
Handle unsupported region

JayKickliter:
`miner_lora_throttle` is not handling the case of when the current region is `undefined` (AKA not yet know), leading to a match failure. This PR crates a new throttle pattern defined as `'unsupported'` and all calls with that handle are no-ops. When the region is callback finally fires in `miner_lora`, the `'unsupported'` handle is replaced with one that does proper throttling.

### DIFF
--- a/src/miner_lora_throttle.erl
+++ b/src/miner_lora_throttle.erl
@@ -27,8 +27,7 @@
     frequency :: number()
 }).
 
--type region() ::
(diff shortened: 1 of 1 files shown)


JayKickliter:
Sorry for the diff noise. I ran whatsapp's erlfmt on the code (which I had been doing before). I can back that out if desired.

Vagabond (approved):

PR #547, opened 2020-10-22
Integration test for transfer hotspot

mrallen1:
Problem to solve: Use the new transfer hotspot transaction to transfer ownership of a gateway in a test. (core PR https://github.com/helium/blockchain-core/pull/659, proto PR https://github.com/helium/proto/pull/54)


mrallen1:
Closing in favor of CT in core

PR #550, opened 2020-10-24
Add EU868 region for UA

amirhaleem:
Currently undefined, meaning that Hotspots in the Ukraine cannot participate in PoC or anything LoRa related. This adds EU868 as the correct region.


abhay (approved):
LGTM

madninja (approved):
This is correct afaict per the Lorawan regional parameter spec

PR #553, opened 2020-10-29
Test ledger sc is disputed when server overspends

vihu:
- Change `client_reports_overspend_test` to `server_doesnt_prevent_overspend_test`
- Explicitly tell router (sc server) to not prevent any overspend (configured via app env)

Intention is to hit at least one code path for ending up with a disputed state channel on the ledger. The other being that there is a causal conflict, this does _not_ do that.


evanmcc:
closing, core

PR #559, opened 2020-11-09
Create New Overlay for Docker and Put Logs in /var/data

lthiery:
The reason for this PR is to make it easy to pull logs from Dockers. 

In theory, the Docker user could just mount `/var/logs/miner` somewhere. Instead, we change `sys.config` to put logs in `/var/data/log` where users are instructed already to mount. This makes it so that current user behavior is unchanged but we benefit by having the logs be more accessible.

Tested this with a local Docker build and it works as intended.


evanmcc (changes requested):
the right way to do this is to do an overlay var in a new profile (in rebar.config) see dev.config and the dev profile for examples.

then you should be able to change the line to:
```RUN rebar3 as docker, prod tar```

and get something sensible, rather than having to use `sed` in the Dockerfile.

lthiery:
> the right way to do this is to do an overlay var in a new profile (in rebar.config) see dev.config and the dev profile for examples.
> 
> then you should be able to change the line to:
> `RUN rebar3 as docker, prod tar`
> 
> and get something sensible, rather than having to use `sed` in the Dockerfile.

That's probably a very fair idea at this point, yeah.

lthiery:
One downside of a standalone config for Docker is that we would need to bump snaps in two spots henceforth. The `sed` route has its benefits.

evanmcc:
you don't?  you just don't override it in your docker config.

lthiery:
Oh - derp. I didn't study the examples enough to know how the overriding worked. Will proceed with the suggestion.

lthiery:
@evanmcc does this look right?

I tested a build for Docker AMD64 and it works as intended. I was able to connect a radio and I saw the logs in `/var/data/log`

lthiery:
@evanmcc could we try to land this in time for the next GA?

 This overlay configuration is actually going to help us with 3rd parties i think, because they can bind the overlay file to a custom `docker.config` for their systems. For example, they can remove the override of the ecc608 configuration to re-enable the ecc608.

evanmcc (approved):

PR #577, opened 2020-12-08
Only allow one p2p address per layer

Vagabond:
In certain conditions we've seen that a single Hotspot is able to submit more
than one PoC receipts and witnesses per PoC layer (hop). While these receipts
are (correctly) filtered later, this PR fixes the underlying condition and and
only allows distinct P2P peers to report a witness/receipt per layer. This
should addresses this rare but concerning behaviour we've observed.

### DIFF
--- a/src/poc/miner_poc_handler.erl
+++ b/src/poc/miner_poc_handler.erl
@@ -26,7 +26,9 @@
     send/2
 ]).
 
--record(state, {}).
+-include("miner_util.hrl").
+
+-record(state, {peer :: undefined | libp2p_crypto:pubkey_bin()}).
 
 %% ------------------------------------------------------------------
 %% API Function Definitions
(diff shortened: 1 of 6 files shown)


src/poc/miner_poc_handler.erl
```diff
@@ -26,7 +26,9 @@
     send/2
 ]).
 
--record(state, {}).
+-include("miner_util.hrl").
+
+-record(state, {peer}).
```
vihu, line 31:
```suggestion
-record(state, {peer :: libp2p_crypto:pubkey_bin()}).
```

src/poc/miner_poc_statem.erl
```diff
@@ -728,8 +728,9 @@ submit_receipts(#data{address=Challenger,
     OnionKeyHash = crypto:hash(sha256, libp2p_crypto:pubkey_to_bin(OnionCompactKey)),
     Path1 = lists:foldl(
         fun({Challengee, LayerHash}, Acc) ->
-            Receipt = maps:get(Challengee, Responses0, undefined),
-            Witnesses = maps:get(LayerHash, Responses0, []),
+            {Address, Receipt} = maps:get(Challengee, Responses0, undefined),
```
vihu, line 731: This will badmatch when `undefined`, case over it maybe?

vihu: added a590c31

src/poc/miner_poc_handler.erl
vihu: added a590c31

PR #580, opened 2020-12-14
Update spreading factor to account for no lorawan header

amirhaleem:
Because PoC packets do not use the LoRaWAN protocol, we don't need to count those 13 bytes (or more) in the spreading factor calculations.


vihu (approved):
nice catch

PR #592, opened 2021-01-06
Fix poc statem to disallow self witnessing

vihu:
This stops including a witness which is the challengee itself at the time of poc receipt txn construction once all the layer responses have been gathered.

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -267,8 +267,8 @@ receiving(info, {blockchain_event, {add_block, _Hash, _, _}}, #data{receiving_ti
     lager:info("got block ~p decreasing timeout", [_Hash]),
     {keep_state, save_data(Data#data{receiving_timeout=T-1})};
 receiving(cast, {witness, Address, Witness}, #data{responses=Responses0,
-                                          packet_hashes=PacketHashes,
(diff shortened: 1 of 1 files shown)


evanmcc:
wouldn't this belong in the receipt validation behind a chain var?

vihu:
> wouldn't this belong in the receipt validation behind a chain var?

Yeah, I have this open on core as well https://github.com/helium/blockchain-core/pull/707 figured having one in miner would also help

src/poc/miner_poc_statem.erl
```diff
@@ -793,6 +799,20 @@ send_onion(P2P, Onion, Retry) ->
             send_onion(P2P, Onion, Retry-1)
     end.
 
+is_self_witness(WitnessPubkeyBin, Challengees, Ledger) ->
+    %% We check for the following:
+    %% - Are we beacon only?
+    %% - If yes, is there are single challengee? (with beacon only, there can only be ONE challengee)
+    %% - If yes, is the challengee the same as this witness, i.e. self-witnessing
+    case blockchain:config(?poc_path_limit, Ledger) of
+        {ok, 1} ->
```
evanmcc, line 808: I'm not sure I like tying these together? maybe just something like `allow_self_witness`?

evanmcc (approved):

PR #595, opened 2021-01-12
Sibyl integration with miner

andymck:
This is the top level branch to expose the new grpc validation service within a miner.  The grpc service is running via the newly added and included app sibyl.  The PR implements the changes to pull sibyl into miner as an included app.

NOTE 1:  the Sibyl app will be pulled into miner by default with this PR, but sibyl will not be started unless the miner env var 'mode' has a value of 'validator' ( as set via Evan's validator PR )


config/sys.config.src
```diff
@@ -0,0 +1,123 @@
+%% -*- erlang -*-
```
evanmcc, line 1: what is this?

andymck: @evanmcc Ignore that added by mistake.  Removed already but I must not have pushed.

rebar.config
```diff
@@ -31,6 +33,7 @@
               ]}.
 
 {plugins, [
+           {grpcbox_plugin, {git, "https://github.com/andymck/grpcbox_plugin.git", {branch, "andymck/ts-master/combined-opts-and-template-changes"}}},
```
evanmcc, line 36: this should just all get pulled in with sibyl, no?

src/miner.app.src
```diff
@@ -26,9 +26,11 @@
     xmerl,  %% nat_upnp has a missing dep
     jsx,
     kvc,
-    longfi
+    longfi,
+    grpcbox,
+    grpc_lib
    ]},
-  {included_applications, [blockchain]},
+  {included_applications, [blockchain, sibyl]},
```
evanmcc, line 33: are we starting this as an included app?  is there a way it can just be part of the regular supervision tree?

src/miner.app.src
```diff
@@ -26,9 +26,11 @@
     xmerl,  %% nat_upnp has a missing dep
     jsx,
     kvc,
-    longfi
+    longfi,
```
evanmcc, line 29: I think that these should be sibyl deps and thus started automatically?

evanmcc (commented):
The more I think about this, the more I feel:
1. Sibyl should be an optional blockchain component, so all this should have here is a check that says "are we a validator?  if so `application:set_env(blockchain, run_sibyl, true)` before starting the blockchain.
2. that Sibyl should be able to force all the generation itself, and not be cluttering up makefiles and configs in applications its included in.  Let me know if you need any help here, I can also pull in Fred and Tristan if anything tricky needs to happen.

src/miner.app.src
andymck: Sibyl is defined as an included app and its sup is started as part of miner_restart_sup but only if the miner env var 'mode' is set to validator

rebar.config
andymck: yeah actually you are right, will remove

src/miner.app.src
andymck: yeah, these are unnecessary.  will remove

andymck:
> The more I think about this, the more I feel:
> 
> 1. Sibyl should be an optional blockchain component, so all this should have here is a check that says "are we a validator?  if so `application:set_env(blockchain, run_sibyl, true)` before starting the blockchain.
> 2. that Sibyl should be able to force all the generation itself, and not be cluttering up makefiles and configs in applications its included in.  Let me know if you need any help here, I can also pull in Fred and Tristan if anything tricky needs to happen.


I agree the suggested approach would be preferred but I did try to do so at one point way back but was unable to make it all work fully.  IIRC the problem was getting the grpcbox plugin to generate the services/behaviours, I could only get this to fire from the top level makefile.  

I can give it another go, back then I was under pressure to get things working and deployed so went with the current approach.  Will revisit.

evanmcc (approved):

PR #597, opened 2021-01-12
Dynamically select PoC TX power by region

amirhaleem:
Set the TX power correctly for US915 and EU868 PoC packets


abhay:
WARNING: DO NOT MERGE THIS WITHOUT SOME PROPER COMMUNICATION

maco2035:
Quick question, This will only set the region if selected right? And not auto switch it to another region unless you change the region in a file, or cli command right?

JayKickliter (approved):

PR #598, opened 2021-01-13
Make onboarding key pubkey

### LINKED ISSUE #594: Update miner to return swarm key as onboarding key
Update to return (and save) swarm key as onboarding key. This should make gateway-config pick it up as the onboarding key

madninja:
This changes makes the onboarding key the same as the public key of the hotspot, but only if there is not already an onboarding key present.

Fixes #594

### DIFF
--- a/src/miner_keys.erl
+++ b/src/miner_keys.erl
@@ -74,7 +74,7 @@ keys({ecc, Props}) when is_list(Props) ->
     OnboardingKey = 
         case ecc508:genkey(ECCPid, public, OnboardingKeySlot) of
             {ok, Key} -> 
-                {ecc_copmact, Key};
+                {ecc_compact, Key};
             {error, ecc_response_exec_error} -> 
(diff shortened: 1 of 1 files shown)


src/miner_keys.erl
```diff
@@ -70,7 +71,15 @@ keys({ecc, Props}) when is_list(Props) ->
                            miner_ecc_worker:get_pid()
                    end,
     {ok, PubKey, KeySlot} = get_public_key(ECCPid, KeySlot0),
-    {ok, OnboardingKey} = ecc508:genkey(ECCPid, public, OnboardingKeySlot),
+    OnboardingKey = 
+        case ecc508:genkey(ECCPid, public, OnboardingKeySlot) of
+            {ok, Key} -> 
+                {ecc_copmact, Key};
```
vihu, line 77:
```suggestion
                {ecc_compact, Key};
```

vihu (approved):

PR #610, opened 2021-01-31
onboarding improvements to CLI and internal miner key management.

ke6jjj:
# Summary

This change adds a new "onboarding" query to the "info" CLI, allowing a user (or script) to obtain the miner's manufacturing and maker details from the new Decentralized Wireless Alliance server.

# Details

This feature fixes a gap in the _manual_ (or some might say _curmudgeonly_) onboarding and location assertion process that appeared as a result of HIP 19 and chain variable transaction number 53, in which the approval of third-party miner manufacturers meant that there would no longer be a single "staking" address to use when constructing `assert_gateway` and `assert_location` transactions.

## Examples

This adds a new `onboarding` sub-command of the `info` command with the following usage:

```
Usage: info onboarding [-k <key>] [-p]

  Get the manufacturing information for this miner, querying the
  Decentralized Wireless Alliance server.

  Options

    -k, --key <onboarding-key>
        Use the specified onboarding key rather than the key for this miner.
    -p, --payer
        After obtaining the information, just print the manufacturer's payer key
        This is the key that should be named when requesting signatures for
        transactions that require manufacturer staking.
 ```

1. Querying all available manufacturing data for this miner.
    _This output is mocked up -- not real data_
    ```
    $ miner info onboarding

    ********************
    General Manufacturing Info
    ********************
    
    +-------------+---------------------------------------------------+
    |    name     |                      result                       |
    +-------------+---------------------------------------------------+
    |    batch    |                       cn                          |
    |  createdAt  |             2021-02-29T00:00:00.000Z              |
    |heliumSerial |            HS1-29-Feb-2021-1303945368             |
    |     id      |                      50686                        |
    |   macEth0   |                c3:a4:4b:2a:e5:15                  |
    |  macWlan0   |                c3:a4:4b:2a:e5:16                  |
    |   makerId   |                        1                          |
    |onboardingKey|         beef2a7df6ddc5fe2f28e8d51de9dead          |          
    |publicAddress|14ADq61qfLrxzx4twXwN3p1RYGa3XWAhyppLHFdGCYBd5jL66v5|
    |  rpiSerial  |                 0000052831786198                  |
    |  updatedAt  |             2020-06-31T00:00:00.000Z              |
   +-------------+---------------------------------------------------+


    ********************
    Manufacturer Info
    ********************

    +------------------+---------------------------------------------------+
    |       name       |                      result                       |
    +------------------+---------------------------------------------------+
    |     address      |13daGGWvDQyTyHFDCPz8zDSVTWgPNNfJ4oh31Teec4TRWfjMx53|
    |    createdAt     |             2021-01-13T01:57:01.662Z              |
    |        id        |                         1                         |
    |locationNonceLimit|                         2                         |
    |       name       |                    Helium Inc                     |
    |    updatedAt     |             2021-01-13T01:57:01.662Z              |
    +------------------+---------------------------------------------------+
    ```

2. Retrieving just the manufacturer's onboarding staking key
    ```
    $ miner info onboarding -p
    13daGGWvDQyTyHFDCPz8zDSVTWgPNNfJ4oh31Teec4TRWfjMx53
    ```

# Testing

I have tested this code on the following miner types:
* `sidetable_v1`
* `sidetable_v2`
* `rakspot`
* `rakspot_v2`

### CODE REVIEW

File: src/cli/miner_cli_info.erl [hunk 1]
```diff
@@ -14,6 +14,7 @@
 
 -define(DEFAULT_LOG_HIT_COUNT, "5").                                                        %% the default number of error log hits to return as part of info summary
 -define(DEFAULT_LOG_SCAN_RANGE, "500").                                                     %% the default number of error log entries to scan as part of info summary
+-define(ONBOARDING_API_URL_BASE, "https://onboarding.dewi.org/api/v2").
```
lthiery, line 17: I'm not sure if you checked in with others on this, but I would imagine the idea of having the Miner depend on ETL for anything is probably not desirable.

ke6jjj: I agree am happy to report that this change doesn't make anything in the miner depend on this server. It's a convenience function in the miner CLI to allow a user or script to determine the maker "address" so that they can construct a proper `assert_gateway` or `assert_location` transaction correctly when using the `txn` CLI verb.

I first proposed having this feature be a part of the wallet CLI, but @madninja hinted that it should be in the miner CLI. So here it is.

ke6jjj: And just to be super clear, this endpoint is only queried when a user asks the miner to do so from the command line. It is not queried during the miner boot process nor consulted at all during regular mining tasks.

ke6jjj:
There were two different tasks being served in this PR. One was the new query feature and the other was to fix a formatting error in PoC logging. @madninja requested that I reduce the changes in this PR to just the task of adding the query, so I have reverted the logging changes and will request them via a separate PR.

ke6jjj:
I just fixed a Dialyzer warning wrt. the return type of the internal helper function `onboarding_info_for_key`. With this change everything passes `make typecheck`.

ke6jjj:
I've just had a volunteer test this code out on a `rakspot_v2` miner as well, and everything works just fine.

src/cli/miner_cli_info.erl
```diff
@@ -229,13 +231,106 @@ info_region(["info", "region"], [], []) ->
...
+        clique_status:table(lists:map(fun tuple_to_table_row/1, MakerDisplay))
+    ].
+
 %%
 %% info summary
 %%
 
 info_summary_cmd() ->
     [
-     [["info", "summary"], [], [], fun info_summary/3],
```
madninja, line 238: perhaps this is the GH diff display, but why was this line removed?

src/miner_keys.erl
```diff
@@ -100,40 +106,30 @@ keys({ecc, Props}) when is_list(Props) ->
...
-                      OnboardingFilename = get_onboarding_filename(),
-                      case filelib:is_file(OnboardingFilename) of
-                          true ->
-                              string:trim(binary_to_list(Key));
-                          false ->
-                              libp2p_crypto:pubkey_to_b58(Key)
-                      end
-               end,
     Props = [{pubkey, MaybeB58(PubKey)},
-             {onboarding_key, MaybeUUIDv4(OnboardingKey)}
```
madninja, line 136: I don't think you can remove MaybeUUIDv4. There' s s set of "blackspots" out there that use UUIDs as their onboarding keys

src/miner_keys.erl
```diff
@@ -22,7 +25,7 @@ get_onboarding_key(Default) ->
         undefined -> Default;
         OnboardingKey ->
             case file:read_file(OnboardingKey) of
-                {ok, Bin} -> Bin;
+                {ok, Bin} -> string:trim(binary_to_list(Bin));
```
madninja, line 28: Could you add a spec to this get_onboarding_key function to make it obvious what it returns?

madninja (changes requested):

src/cli/miner_cli_info.erl
ke6jjj: This line wasn't doing anything. It essentially specifies that you can run `info summary` without any additional arguments. However, the line below it specifies that you can run `info summary` without any additional arguments _or_ with the flag `-e|--error_count`.

src/miner_keys.erl
ke6jjj: This code doesn't remove the ability for an onboarding key to be a UUIDv4. Instead, it shifts the responsibility of deciding whether it is a UUIDv4 or B58 to the place where it belongs: in the code that knows whether it is a UUIDv4 or a B58 key! Once the code reaches this point, `OnboardingKey` will already be in whatever format it needs to be in for dumping.

src/miner_keys.erl
ke6jjj: Certainly!

src/miner_keys.erl
ke6jjj: And when I said:

>  ... in the code that knows whether it is a UUIDv4 or a B58 key!

The code that I was speaking of is `keys()/1`,

src/miner_keys.erl
ke6jjj: Done.

src/miner_keys.erl
madninja: Right, sorry, my point is that `keys/1` is somehow _not_ handling the fact that an onboarding key can be a uuid?

ke6jjj: Nope. I think the point will be best illustrated when we revisit what `MaybeUUIDv4` _was_ before I deleted it. It was badly named. This is what it did:

```
MaybeUUIDv4 = fun(undefined) -> undefined;		
                   (Key) ->		
                       %% only production blackspots read onboarding_key from a file       		
                       %% and they're (unfortunately) considered to be uuids		
                       OnboardingFilename = get_onboarding_filename(),		
                       case filelib:is_file(OnboardingFilename) of		
                           true ->		
                               string:trim(binary_to_list(Key));		
                           false ->		
                               libp2p_crypto:pubkey_to_b58(Key)		
                       end		
                end,
```

This has nothing to do with supporting UUIDv4 strings and everything to do with duplicating the logic of finding out where the onboarding key came from and _then_ turning it into a string.

madninja: It's been a while, but I think I follow.. This code may not have been doing anything useful in practice, I guess. I just want to make sure that it doesn't fall over because the onboarding key is s a uuidv4 and keys/1 now does a pubkey_to_b58 by default. 

Perhaps get_onboarding_key/1 is the place where uuid onboarding keys are handled

ke6jjj: We just crossed comment paths. I updated my comment while you were commenting. The additions might make it more clear.

ke6jjj: > I just want to make sure that it doesn't fall over because the onboarding key is s a uuidv4 and keys/1 now does a pubkey_to_b58 by default.

I tested this code on a rak v1, which if you recall, really is a `blackspot`. The onboarding key there is a UUIDv4, which comes from the filesystem. `keys({file, ...})`. This causes `keys({file, ...})` to be run, which "does the right thing".

ke6jjj:
Is this stalled for any particular reason?

src/miner_keys.erl
ke6jjj: Resolving as I think we're in agreement.

PR #613, opened 2021-02-02
Update poc channel for EU

amirhaleem:
Allows for 500mW PoC transmissions in EU868. Must be merged with https://github.com/helium/sx1302_hal/pull/13

### DIFF
--- a/src/poc/miner_onion_server.erl
+++ b/src/poc/miner_onion_server.erl
@@ -344,8 +344,8 @@ decrypt(Type, IV, OnionCompactKey, Tag, CipherText, RSSI, SNR, Frequency, Channe
             case miner_lora:location_ok() of
                 true ->
                     %% the fun below will be executed by miner_lora:send and supplied with the localised lists of channels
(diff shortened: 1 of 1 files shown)


Vagabond:
I think this will break the PoC "channel" stuff:

https://github.com/helium/blockchain-core/blob/master/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl#L326-L329

as we assume PoC packets are uniformly distributed across the 8 uplink channels right now.

amirhaleem:
I see the issue. not sure how best to solve this without exposing region in core?

mfalkvidd:
This suggestion would force all LoRaWAN devices on the network to use a single channel. That would have a huge impact on the overall capacity.

If LoRaWAN devices do not use this single channel (and special batteries that can support the 500mW discharge rate, most Saft batteries do not support that), PoC in EU would become very artificial, since it uses much more powers than devices will.

timcooijmans:
Please don't do this: As @mfalkvidd mentions, it will give a unrealistic PoC indication while actual coverage will be much less (being limited by 25mW instead of 500mW).

PR #615, opened 2021-02-05
Switch blockchain for testing NAT fixes

vihu:


rebar.config
```diff
@@ -10,7 +10,7 @@
 
 {deps, [
         {blockchain, {git, "https://github.com/helium/blockchain-core.git",
-                      {branch, "master"}}},
+                      {branch, "rg/nat-fixes"}}},
```
vihu, line 13: Switch this back to master when https://github.com/helium/blockchain-core/pull/728 lands

PR #619, opened 2021-02-08
Useablity improvments to miner info commands

maco2035:
There was missing "info" in the display. Made it look like the other commands such as peer.
Also made it more clear for what info_in_consensus_usage(), and info_name_usage() does.
Added the region to the info_usage(). Was missing in the past.

These are all just small changes, no changes to the base codebase, only to the UI of the CLI interface for the miners.


evanmcc (approved):

evanmcc:
Thanks!

PR #624, opened 2021-02-10
Update packet_rssi function to use RSSIS if available

amirhaleem:
SX1302's use the `rssis` field from the packet forwarder to denote the strength of the signal, versus `rssi` which is the strength of the *channel*. SX1302 `rssis` appears to be behave like SX1301 `rssi`, so this update uses `rssis` if available.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -694,7 +694,8 @@ tmst_to_local_monotonic_time(Tmst_us, PrevTmst_us, PrevMonoTime_us) ->
 %% GWMP JSON V1/V2.
 -spec packet_rssi(map()) -> number().
 packet_rssi(Packet) ->
-    case maps:get(<<"rssi">>, Packet, undefined) of
+    %% SX1302 uses RSSIS and SX1301 uses RSSI
+    case maps:get(<<"rssis">>, Packet, maps:get(<<"rssi">>, Packet, undefined)) of
         %% GWMP V2
         undefined ->
             %% `rsig` is a list. It can contain more than one signal


Vagabond (approved):
Don't merge this until Jay approves though

lthiery:
We covered this internally on the Clubhouse ticket: https://app.clubhouse.io/hlm/story/8886/miner-should-tolerate-v2-of-gwmp

Essentially, `rrsi` in GWMP v1 corresponds to `rssic` in GMPW v2. If you choose `rssis` from GWMP v2, your data between packet forwarders that use GWMP v1 and v2 won't be homogenous.

[This document](https://www.thethingsnetwork.org/forum/uploads/default/original/1X/4fbda86583605f4aa24dcedaab874ca5a1572825.pdf) explains: 

```
The difference between the RSSI channel and RSSI signal is observable at low input signal power: the RSSI channel is equal to the noise floor whereas the RSSI signal is equal to the input signal. The RSSI signal stays linear over the range of the input signal, in contrary to the RSSI channel.

In the JSON protocol for gateway V1, only the rssi field is present as the SX1301 only provides the RSSI channel value, not the RSSI signal one. In the gateway V2, the RSSI signal is provided by the DSP present on board. The JSON protocol includes an additional field rsig which contains both fields; rssic (RSSI channel) and rssis (RSSI Signal).
```

JayKickliter (approved):

amirhaleem:
> Essentially, `rssi` in GWMP v1 corresponds to `rssic` in GMPW v2. If you choose `rssis` from GWMP v2, your data between packet forwarders that use GWMP v1 and v2 won't be homogenous.

is the 1302_hal using GWMP v2 or v1?

lthiery:
It appears like the sx1302_hal has some extended version of GWMP JSON v1, where they added `rssic`.

`rxpk.rssis` is provided:
https://github.com/helium/sx1302_hal/blob/helium/hotspot/packet_forwarder/src/lora_pkt_fwd.c#L2026

and so is `rxpk.rssi`, ie `rssic`:
https://github.com/helium/sx1302_hal/blob/helium/hotspot/packet_forwarder/src/lora_pkt_fwd.c#L2069

GWMP JSON v2 though specifies `rssic`/`rssis` as fields of the `rsig`, which exist as an array in the top-level `rxpk` object

```
rxpk
rsig    | object | array of object Received signal information, per antenna 
```

Anyway, looks like its merged so I guess its unimportant.

PR #625, opened 2021-02-10
Update packet_rssi function to use RSSIS from GWMP JSON v2

lthiery:
Since #624 emphasizes using RSSIS whenever available, we should be doing it for GWMP JSON v2 as well.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -704,9 +704,9 @@ packet_rssi(Packet) ->
             %% highest RSSI[Channel]
             [H|T] = maps:get(<<"rsig">>, Packet),
             Selector = fun(Obj, Best) ->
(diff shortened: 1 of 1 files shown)


abhay:
This affects miners in the world using a DIY that speaks GWMP v2.

lthiery:
Just to clarify the idea here: with #624 we are saying that packet_rssi returns `rssis` when `rssis` is available; otherwise, return `rssic`. However, although we implemented that changes for GWMP Json V1 frames, we left GWMP Json V2 frames on `rssic`. This PR makes our logic of favoring `rssis` consistent among GWMP Json V1 and V2.

ke6jjj:
I feel this has been implemented already with the PoC v11 activation. It's probably time to close.

PR #629, opened 2021-02-11
Validators

evanmcc:
This is the main PR for the validators feature described in HIP 25: https://github.com/helium/HIP/blob/master/0025-validators.md

It is not entirely feature complete, but is close enough to start review.

Depends on: https://github.com/helium/blockchain-core/pull/727 and https://github.com/helium/proto/pull/59


nbahramzad:
As someone that would like to run a Validator on an arm64 server, could you please include the appropriate Dockerfile and build pipeline files?

src/miner_val_heartbeat.erl
```diff
@@ -0,0 +1,118 @@
...
+                            case blockchain_ledger_validator_v1:last_heartbeat(Val) of
+                                N when (N + HBInterval) =< Height andalso (not Sync) -> 
+                                    %% we need to construct and submit a heartbeat txn
+                                    {ok, CBMod} = blockchain_ledger_v1:config(?predicate_callback_mod, Ledger),
+                                    {ok, Callback} = blockchain_ledger_v1:config(?predicate_callback_fun, Ledger),
+                                    UnsignedTxn =
+                                        blockchain_txn_validator_heartbeat_v1:new(Address, Height, CBMod:Callback()),
+                                    Txn = blockchain_txn_validator_heartbeat_v1:sign(UnsignedTxn, SigFun),
+                                    lager:debug("submitting txn ~p for val ~p ~p ~p", [Txn, Val, N, HBInterval]),
+                                    blockchain_worker:submit_txn(Txn),
```
vihu, line 81: Just wondering if the fire and forget here for the heartbeat txn is sufficient enough, do we need to do `submit/2`, supply a callback and resend if the heartbeat txn fails say X number of times?

test/miner_blockchain_SUITE.erl
```diff
@@ -22,6 +22,9 @@
 all() -> [
           restart_test,
           dkg_restart_test,
+          validator_test,
```
vihu, line 25: I'm only nitpicking but can we have an entirely separate suite with the three tests here if it's not too much pain? `miner_validator_SUITE` perhaps?

src/miner_val_heartbeat.erl
Vagabond: This loop might spam a lot of redundant transactions as-is, it might be better to use the callback to wait for the one submitted to clear/fail instead of trying every single block.

andymck:
lgtm.  On a related note, i will use the added 'mode' env var to determine whether to start sibyl in miner

src/cli/miner_cli_genesis.erl
```diff
@@ -191,8 +192,10 @@ forge(PubKeyB58, ProofB58, Addrs, N, Curve) ->
 
 
     LocAddresses = lists:zip(Locations, Addresses),
-    InitialGatewayTransactions = [blockchain_txn_gen_gateway_v1:new(Addr, Addr, Loc, 0)
-                                  || {Loc, Addr} <- LocAddresses ],
+    %% InitialGatewayTransactions = [blockchain_txn_gen_gateway_v1:new(Addr, Addr, Loc, 0)
```
Vagabond, line 195: Remove?

src/cli/miner_cli_genesis.erl
```diff
@@ -389,5 +392,11 @@ make_vars() ->
       ?snapshot_version => 1,
       ?sc_grace_blocks => 20,
       ?snapshot_interval => 5,
-      ?rewards_txn_version => 2
+      ?rewards_txn_version => 2,
+      ?validator_version => 2,
```
Vagabond, line 396: This should probably be 3 once we merge the intermediate unstake version?

src/miner.app.src
```diff
@@ -7,6 +7,7 @@
    [
     kernel,
     stdlib,
+    sasl,
```
Vagabond, line 10: Why?

src/miner.erl
```diff
@@ -727,7 +733,7 @@ set_next_block_timer(State=#state{blockchain=Chain}) ->
 %% final 0.04 (twice as much leverage, but 20% of the rate).
 
 %% when drift is small or 0, let it accumulate for a bit
-catchup_time(N) when N < 0.0001 ->
+catchup_time(N) when N < 0.001 ->
```
Vagabond, line 736: This intentional?

src/miner_util.erl
```diff
@@ -87,26 +87,34 @@ mark(Module, Mark) ->
...
-                Map#{<<"gps_fix_quality">> => <<"bad_assert">>};
-            %% no gps fix
-            {error, no_fix} ->
-                Map#{<<"gps_fix_quality">> => <<"no_fix">>};
-            %% no location asserted somehow
-            {error, not_asserted} ->
-                Map#{<<"gps_fix_quality">> => <<"not_asserted">>};
-            %% got nonsense or a hopefully transient error, return nothing
+        case application:get_env(miner, mode, gateway) of
+            validator ->
```
Vagabond, line 91: Since we don't use the gps information at all, can we just always send the 'release version'?

src/miner_consensus_mgr.erl
```diff
@@ -134,7 +133,19 @@ group_predicate("consensus" ++ _ = Name) ->
                     lager:debug("con pred ~p eh ~p h ~p", [Name, Height0, ElectionHeight]),
                     Height = list_to_integer(Height0),
                     Height < ElectionHeight;
-                [_Tag, Height0, _Hash] ->
+                _ ->
```
Vagabond, line 136: Remind me why this changed?

Vagabond: Are these just no longer possible?

src/miner_val_heartbeat.erl
```diff
@@ -0,0 +1,129 @@
...
+handle_call(_Request, _From, State) ->
+    lager:warning("unexpected call ~p from ~p", [_Request, _From]),
+    Reply = ok,
+    {reply, Reply, State}.
+
+handle_cast(_Msg, State) ->
+    lager:warning("unexpected cast ~p", [_Msg]),
+    {noreply, State}.
+
+handle_info({blockchain_event, {add_block, _Hash, _Sync, _Ledger}},
```
Vagabond, line 61: What is the purpose of the waiting state? just to avoid doing extra checks for a few blocks?

src/miner_val_heartbeat.erl
```diff
@@ -0,0 +1,129 @@
...
+                    lager:debug("getting validator for address ~p not found", [Address]),
+                    {noreply, State}
+            end;
+        _ ->
+            {noreply, State}
+    end;
+%% logically it doesn't matter what the result is, we either need to start actively waiting or
+%% trying again
+handle_info({sub, _Res}, State) ->
+    {noreply, State#state{txn_status = ready, txn_wait = 10}};
```
Vagabond, line 105: Would it not be better to look at the result here and also use the `validator_liveness_interval` to inform the txn_wait?

test/miner_blockchain_SUITE.erl
Vagabond: Yes, we should have these more broken up, but that's probably out of scope for this right now.

test/miner_blockchain_SUITE.erl
```diff
@@ -216,6 +244,310 @@ dkg_restart_test(Config) ->
...
+
+    UTxn2 = blockchain_txn_transfer_validator_stake_v1:new(Addr1, Addr2, Owner, ?bones(100000)),
+    Txn2 = blockchain_txn_transfer_validator_stake_v1:sign(UTxn2, AuxSigFun),
+    ct:pal("submitting ~p", [Txn2]),
+    _ = [ok = ct_rpc:call(M, blockchain_worker, submit_txn, [Txn2]) || M <- Miners1],
+
+    %% transfer the stake from new node 1 to new node 2 and confirm that 1 leaves and 2 joins
+    Miners2 = [NewVal2 | Miners],
+    AddrList2 = [{NewVal2, Addr2} | AddrList1],
+    receive _ -> receive _ -> ok after 0 -> ok end after 0 -> ok end,
```
Vagabond, line 315: Are these the transactions clearing??

rebar.config
```diff
@@ -10,7 +10,7 @@
 
 {deps, [
         {blockchain, {git, "https://github.com/helium/blockchain-core.git",
-                      {branch, "master"}}},
+                      {branch, "pevm/testnet-integration1"}}},
```
vihu, line 13: reminder, switch to master when branch lands. also looks like no corresponding PR on core side?

src/handlers/miner_dkg_penalty_handler.erl
```diff
@@ -0,0 +1,232 @@
...
+                          lager:debug("adding sig from conf"),
+                          [{Address, Signature}|Acc];
+                      {true, false} ->
+                          lager:debug("got invalid signature ~p from ~p", [Signature, Address]),
+                          Acc;
+                      {false, _} ->
+                          Acc
+                  end
+          end, State#state.signatures, Signatures),
+        Majority = floor(length(State#state.members) / 2) + 1,
```
vihu, line 97: nit: indentation

src/handlers/miner_dkg_penalty_handler.erl
```diff
@@ -0,0 +1,232 @@
...
+    case length(GoodSignatures) of
+        SigLen when SigLen >= (State#state.f*2) + 1 ->
+            done(GoodSignatures, State#state.artifact),
+            {State#state{signatures=GoodSignatures, done_called=true}, []};
+        SigLen when SigLen >= Majority andalso State#state.conf_sent == false ->
```
vihu, line 102: I'm wondering whether this node should multicast `{conf, GoodSigs}` even if it sent conf before when it sees `SigLen >= Majority`, mostly because it would help propagate conf msgs faster, I _think_. If I'm entirely mistaken, can we add a little comment here with some reasoning plz?

src/handlers/miner_dkg_penalty_handler.erl
```diff
@@ -0,0 +1,232 @@
...
+            lager:warning("got truncated message: ~p:", [BinMsg]),
+            ignore
+    end;
+handle_message({conf, Signatures}, _Index, State)
+  when State#state.artifact /= undefined andalso State#state.done_called == false ->
+    GoodSignatures =
+        lists:foldl(
+          fun({Address, Signature}, Acc) ->
+                  %% only check signatures from members we have not already verified and have not already appeared in this list
+                  case {lists:keymember(Address, 1, Acc) == false andalso
```
vihu, line 84: Maybe it doesn't matter but the order of operation here is flipped I think, membership check in state.members should happen first imo

src/handlers/miner_dkg_penalty_handler.erl
```diff
@@ -0,0 +1,232 @@
...
+%%%-------------------------------------------------------------------
+-module(miner_dkg_penalty_handler).
+
+-behavior(relcast).
+
+-export([init/1, handle_message/3, handle_command/2, callback_message/3, serialize/1, deserialize/1, restore/2]).
+
+
+-record(state, {bbas :: #{pos_integer() => hbbft_bba:bba_data()},
+                privkey :: tpke_privkey:privkey(),
```
vihu, line 28: Uhm, is this module missing handling `tc_key_share:tc_key_share()` entirely?

test/miner_blockchain_SUITE.erl
evanmcc: since these are redundant now (they're now tested in core), I'm going to delete all but the transition test.

src/miner.erl
evanmcc: yes

src/miner.app.src
evanmcc: removing

src/miner_util.erl
evanmcc: this is a good idea, but we'd need to change it to read lsb-release on the gateways

src/miner_consensus_mgr.erl
evanmcc: yeah, that was an ancient clause for backwards compatibility that I finally got rid of

src/miner_val_heartbeat.erl
evanmcc: so we don't spam a new txn while waiting for the old one to land

src/miner_val_heartbeat.erl
evanmcc: I don't actually need to look at the response here, since if we succeeded, we won't submit another one, and if we failed, we'll just submit another one in the next block we get

rebar.config
evanmcc: most of the core stuff has already landed, and I think that we have landed all the subsequent relevant PRs as well.  We do need to upgrade, though.

Vagabond (approved):

PR #631, opened 2021-02-11
Adding arm64 support for validator branch.

nbahramzad:
Pretty self-explanatory, just added build pipeline change + Dockerfile for an arm64 validator image.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -36,6 +36,19 @@ steps:
     commands:
       - "git fetch -t"
       - ".buildkite/scripts/make_val_image.sh"
+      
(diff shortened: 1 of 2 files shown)


DGKSK8LIFE (approved):

nbahramzad:
As discussed on #validator-dev Discord, the AMD64 Docker build fails in the same way. Once the correct tag is added to rebar.config, this PR will build correctly.

evanmcc:
Thanks for this.  I don't think that this will work on our infra without pulling some other stuff over for cross compilation.  Let me check with some people and I'll try to get that added today.

evanmcc (approved):

PR #632, opened 2021-02-11
Pevm/validators

rpmarkus:
Build is not happy with {semver, "val"}

throwing this:
Resolved miner-al: No names found, cannot describe anything.+build.fatal:Invalidobjectname'fatal'..

changed to same syntax as prod.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -212,7 +212,7 @@
              [
               {relx, 
                [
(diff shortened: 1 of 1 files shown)


evanmcc:
sorry, this will be resolved shortly, but this isn't the proper fix.

PR #642, opened 2021-02-19
fix validator profile in rebar.config

olb1ue:
This updates the validator profile in rebar.config to create the usual folder structure with the config files. Same structure as the docker profile, but using val.config instead of docker.config.


olb1ue:
This got done!

PR #643, opened 2021-02-19
V2 Sibyl Integration with Miner

andymck:
Provides V2 gateway APIs, including routing and state channels


andymck:
amalgamated into https://github.com/helium/miner/pull/595

Issue #650, opened 2021-02-24
Bump deps


JayKickliter:
The following deps have build-system improvements for macOS builds and OTP23 support

- [ ] https://github.com/helium/erlang-pbc
- [ ] https://github.com/helium/erlang-erasure
- [ ] https://github.com/helium/ebus

These deps have outstanding issues:

- [ ] https://gitlab.com/barrel-db/erlang-rocksdb/-/issues/118

evanmcc:
This is done other than horror related to erocks that will take some time to resolve.

PR #651, opened 2021-02-24
gather up osx+freebsd build changes

evanmcc:
This pulls in a number of changes to various NIF-using subsystems to make them build better on recent OS X and FreeBSD.

I've tested that it works on OS X Big Sur with OTP 22 and OTP 23.


JayKickliter:
This addresses all but the last item (RocksDB) in #650

Vagabond:
Rocksdb is more complicated as we have that crosscompile patch we need to rebase as well.

Vagabond (approved):
Seems to build and run ok here on void linux as well on both OTP 22 and OTP 23.

PR #660, opened 2021-03-01
Update Readme for Erlang OTP 23.

philltran:
Updating Readme for current supported version of Erland OTP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ git clone https://github.com/helium/miner.git
 
 ### Install Erlang
 
-Miner has been tested against Erlang OTP 22.3.1.
+Miner has been tested against Erlang OTP 23.2.3.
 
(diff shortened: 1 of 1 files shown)


Vagabond:
It would be nice to say we support OTP 22 and 23 (and there's just a couple small changes outstanding for OTP 24) rather than specifying an exact version.

philltran:
I updated the PR to read: 

> Miner has been tested against Erlang OTP 22 and 23. Follow [this PR](https://github.com/helium/miner/pull/655) for progress on Erlang OTP 24 support.

philltran:
There is not a raspian version for OTP 23. Should I use the Ubuntu instructions?
https://www.erlang-solutions.com/downloads/

Vagabond:
Ugh, I'm not sure what's best to do there. OTP 22 would be ok as well.

evanmcc (approved):

PR #661, opened 2021-03-02
Fix potential watchdog crash when printing keys

Vagabond:


Vagabond:
I tested this works on a hotspot with an ECC

madninja (approved):

PR #687, opened 2021-03-09
Bump DKG and HBBFT for latest PBC and Erasure, respectively

JayKickliter:
`erlang-pbc` and `erlang-erasure` have recent changes to improve cross compiling OpenWRT targets. 

https://github.com/helium/erlang-pbc/pull/15
https://github.com/helium/erlang-erasure/pull/12


Vagabond (approved):
No objections, we will just need to make sure we test these on the hotspots well before any GA including them.

PR #697, opened 2021-03-16
Decrease addr hash false positive rate and avoid double-adding a block

Vagabond:


src/poc/miner_poc_statem.erl
```diff
@@ -687,7 +687,7 @@ maybe_init_addr_hash(#data{blockchain=Blockchain, addr_hash_filter=#addr_hash_fi
                                     %% ok, everything lines up
                                     Data;
                                 _ ->
-                                    case blockchain:get_block(Height, Blockchain) of
+                                    case blockchain:get_block(Height+1, Blockchain) of
```
JayKickliter, line 690: What's the `+1` for and how did this work before?

Vagabond: It simply added the same hashes for a block twice, which was harmless but unnecessary.

JayKickliter (approved):

Issue #677, opened 2021-03-08
Missing LoraWAN support code frequency for my country, Uganda.


ignited2021:
According to the information provided by Lora alliance org on the list of countries with supported frequencies for LoraWan My country (Uganda) is among and lies within this range of frequency; 923 - 925 MHz plus 863 - 865 MHz,
865-867.6 MHz, 869.25 - 869.7 MHz; with these support codes (EU433 and AS923). I would like to start mining Helium (HNT) in Uganda However I'm not sure whether the frequencies in Uganda will be compatible with the LoRaWAN devices. The Github list does not show whether Uganda frequency ranges support the Helium network in Uganda in case I buy/order/ship my the devices to Uganda and connect them will they work? Plus I see that my country shares the same codes on the Lora alliance org with many other countries like Hong Kong, UK, Thailand, that are currently mining Helium but on your list Uganda's information is not matching that on Lora alliance org. Could you please crosscheck for me the frequency to confirm if I can mine Helium in Uganda using the current mining LoRaWAN devices within those radio frequency ranges in Uganda as per Lora alliance organization list of supported countries.

timtamothee:
Hi @ignited2021, can you please provide the documentation that you found from the LoRa Alliance site for Uganda's supported LoRaWAN frequency for reference? This can help others cross-reference before approving a supported frequency. Thanks.

shawaj:
I believe based on the document linked in https://github.com/helium/miner/pull/663 the correct frequency for Uganda is EU868

jmarcelino:
According to RP002-1.0.1 the LoRa Alliance is recommending IN865 for Uganda and this is supported by the Uganda Communications Commission https://www.ucc.co.ug/wp-content/uploads/2017/09/Technical-Requirements-for-operation-of-868.docx.pdf

From that document you can also see that in Uganda there is a condition that "Any gateways or base stations deployed shall have to be registered with UCC."

shawaj:
@jmarcelino ah yes, the LoRa doc does suggest in865. I must have made an error whilst translating that to the CSV.

However, the UCC document you linked doesn't seem to support the frequency ranges mentioned in the lora doc and seems to allow eu868 as well.

And since Kenya, Rwanda, Burundi, Zambia all use eu868 it seems to make more sense to use eu868 than in865 in this case, IMO.

jmarcelino:
[...]

PR #705, opened 2021-03-23
frequency: update Uganda to correct frequency

shawaj:
Accidentally put EU868 for Uganda instead of IN865 as per the lora doc

Closes: #677 
Change-type: patch
Signed-off-by: Aaron Shaw <shawaj@gmail.com>


shawaj:
@Vagabond as per #677 I made a mistake on Uganda frequency

PR #708, opened 2021-03-26
Check gateway capabilities in POC flows

andymck:
Linked blockchain core PR: https://github.com/helium/blockchain-core/pull/754

Add capability check to the POC flows.  If a GW does not have the required capability for example to send POC requests, then it will be prevented from doing so.

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -41,6 +41,7 @@
 -endif.
 
 -include_lib("blockchain/include/blockchain_vars.hrl").
+-include_lib("blockchain/include/blockchain_caps.hrl").
 
 -define(SERVER, ?MODULE).
 -define(MINING_TIMEOUT, 30).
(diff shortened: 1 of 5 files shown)


src/poc/miner_poc_statem.erl
```diff
@@ -41,6 +41,7 @@
 -endif.
 
 -include_lib("blockchain/include/blockchain_vars.hrl").
```
evanmcc, line 43: can we take care of this in the init instead?  rather than doing all this stuff, just don't even register for events if we're not capable of handling them.

evanmcc (changes requested):
I think that this is ready pending the core side, but I think that it'd be simpler just to not init the poc statem (not even starting it would be better, but hard to do in testing with indeterminite chain states) to follow chain events unless we have the proper capability.

andymck:
> I think that this is ready pending the core side, but I think that it'd be simpler just to not init the poc statem (not even starting it would be better, but hard to do in testing with indeterminite chain states) to follow chain events unless we have the proper capability.

I did consider that approach but as this evolves the light GWs will be allowed to witness.  As such the implemented approach allows for such changes with minimal additions.

evanmcc (approved):

PR #712, opened 2021-03-28
Refactor create_block

xandkar:
Enhancing understandability and Dialyzability. Not done yet.

### DIFF
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -661,12 +661,13 @@ snapshot_hash(Ledger, BlockHeightNext, Metadata, VotesNeeded) ->
     end.
 
 -spec common_enough_or_default(non_neg_integer(), [X], X) -> X.
+common_enough_or_default(_, [], Default) ->
+    Default;
 common_enough_or_default(Threshold, Xs, Default) ->
     %% Looking for highest count AND sufficient agreement:
     case miner_util:list_count_and_sort(Xs) of
         [{X, C}|_] when C >= Threshold -> X;
-        [{_, _}|_]                     -> Default; % Not common-enough.
-        []                             -> Default
+        [{_, _}|_]                     -> Default % Not common-enough.
     end.
 
 set_next_block_timer(State=#state{blockchain=Chain}) ->
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -1,8 +1,3 @@
-%%%-------------------------------------------------------------------
-
-%% @doc miner
-%% @end
-%%%-------------------------------------------------------------------
 -module(miner).
 
 -behavior(gen_server).
@@ -10,9 +5,7 @@
 -include_lib("blockchain/include/blockchain_vars.hrl").
 -include_lib("blockchain/include/blockchain.hrl").
 
-%% ------------------------------------------------------------------
-%% API Function Exports
-%% ------------------------------------------------------------------
+%% API
 -export([
     start_link/0,
     p2p_status/0,
@@ -30,9 +23,13 @@
     version/0
 ]).
 
-%% ------------------------------------------------------------------
-%% gen_server Function Exports
-%% ------------------------------------------------------------------
+-export_type([
+    create_block_ok/0,
+    create_block_error/0,
+    create_block_result/0
+]).
+
+%% gen_server
 -export([
     init/1,
     handle_call/3,
@@ -40,6 +37,42 @@
     handle_info/2,
     terminate/2
 ]).
+
+
+-type metadata_v1() ::
+    {integer(), blockchain_block:hash()}.
+
+-type metadata_v2() ::
+    #{
+        timestamp      => integer(),
+        seen           => binary(),
+        bba_completion => binary(),
+        head_hash      => blockchain_block:hash(),
+        snapshot_hash  => binary()
+     }.
+
+-type metadata() ::
+    [{J :: pos_integer(), M :: metadata_v2() | metadata_v1()}].
+
+-type swarm_keys() ::
+    {libp2p_crypto:pubkey(), libp2p_crypto:sig_fun()}.
+
+-type create_block_error() ::
+      stale_hash
+    | multiple_hashes.
+
+-type create_block_ok() ::
+    #{
+        address               =>  libp2p_crypto:pubkey_bin(),
+        unsigned_binary_block =>  binary(),
+        signature             =>  binary(),
+        pending_txns          =>  blockchain_txn:txns(),
+        invalid_txns          =>  blockchain_txn:txns()
+    }.
+
+-type create_block_result() ::
+      {ok, create_block_ok()}
+    | {error, create_block_error()}.
 
 -record(state, {
     %% NOTE: a miner may or may not participate in consensus
@@ -50,7 +83,7 @@
     current_height = -1 :: integer(),
     blockchain_ref = make_ref() :: reference(),
     swarm_tid :: ets:tid() | atom(),
-    swarm_keys :: {libp2p_crypto:pubkey(), libp2p_crypto:sig_fun()}
+    swarm_keys :: swarm_keys()
 }).
 
 -define(H3_MINIMUM_RESOLUTION, 9).
@@ -86,27 +119,18 @@ inc_tv(Incr) ->
 
 -endif.
 
-%% ------------------------------------------------------------------
-%% API Function Definitions
-%% ------------------------------------------------------------------
+%% ----------------------------------------------------------------------------
+%% API
+%% ----------------------------------------------------------------------------
 
 start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], [{hibernate_after, 5000}]).
 
-%%--------------------------------------------------------------------
-%% @doc
-%% @end
-%%--------------------------------------------------------------------
 block_age() ->
     Chain = blockchain_worker:blockchain(),
     {ok, Block} = blockchain:head_block(Chain),
     erlang:system_time(seconds) - blockchain_block:time(Block).
 
-
-%%--------------------------------------------------------------------
-%% @doc
-%% @end
-%%--------------------------------------------------------------------
 -spec p2p_status() -> [{Check::string(), Result::string()}].
 p2p_status() ->
     SwarmTID = blockchain_swarm:tid(),
@@ -148,10 +172,6 @@ p2p_status() ->
                           {CheckNatType, "nat_type"},
                           {CheckHeight, "height"}]).
 
-%%--------------------------------------------------------------------
-%% @doc
-%% @end
-%%--------------------------------------------------------------------
 %% TODO: spec
 relcast_info(Group) ->
     Mod = case Group of
@@ -166,10 +186,6 @@ relcast_info(Group) ->
             libp2p_group_relcast:info(Pid)
     end.
 
-%%--------------------------------------------------------------------
-%% @doc
-%% @end
-%%--------------------------------------------------------------------
 %% TODO: spec
 relcast_queue(Group) ->
     Mod = case Group of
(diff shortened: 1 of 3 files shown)


src/miner.erl
```diff
@@ -600,9 +455,219 @@ terminate(Reason, _State) ->
...
+    % we expect every stamp to contain the same block hash
+    {ok, Consensus_Members} =
+        blockchain_ledger_v1:consensus_members(blockchain:ledger(Chain)),
+    N = length(Consensus_Members),
+    F = ((N - 1) div 3),
+    {_, Hashes} = meta_to_stamp_hashes(Metadata),
+    case process_hashes(Hashes, F) of
+        [CurrentBlockHash] ->
+            {ok, create_block(Metadata, Txns, HBBFTRound, Chain, N, F, Swarm_Keys)};
+        [_OtherBlockHash] ->
```
Vagabond, line 491: Add a log message on what the other hash was and how many people picked it

src/miner.erl
```diff
@@ -41,16 +38,52 @@
...
+        invalid_txns          =>  blockchain_txn:txns()
+    }.
+
+-type create_block_result() ::
+      {ok, create_block_ok()}
+    | {error, create_block_error()}.
+
 -record(state, {
-    %% NOTE: a miner may or may not participate in consensus
+    % NOTE: a miner may or may not participate in consensus
```
evanmcc, line 78: please don't do this, it messes up auto-formatting.  % is for line end comments %% is for comments with matching indentation

src/miner_util.erl
```diff
@@ -6,17 +6,37 @@
...
+
+%%-----------------------------------------------------------------------------
+%% @doc `list_count` then sort from largest-first (head) to smallest-last.
 %% @end
-%%--------------------------------------------------------------------
+%%-----------------------------------------------------------------------------
+-spec list_count_and_sort([A]) ->
+    [{A, pos_integer()}].
+list_count_and_sort(Xs) ->
+    lists:sort(fun({_, C1}, {_, C2}) -> C1 > C2 end, maps:to_list(list_count(Xs))).
```
evanmcc, line 38: this can also be `lists:keysort(2, ...)`

src/miner.erl
```diff
@@ -600,9 +444,236 @@ terminate(Reason, _State) ->
...
 
-%% ==================================================================
-%% Internal functions
-%% =================================================================
+%% ============================================================================
+%% Private
+%% ============================================================================
+
+-spec try_create_block(
+    metadata(),
```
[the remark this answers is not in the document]
xandkar: There's no type-repetition here - can't save any space with `Name ::`, so it spills either vertically or horizontally - I'd much rather vertically, so I can split-screen.

src/miner.erl
```diff
@@ -600,9 +444,236 @@ terminate(Reason, _State) ->
...
+    non_neg_integer(),
+    blockchain:blockchain(),
+    N :: non_neg_integer(),
+    F :: non_neg_integer(),
+    swarm_keys()
+) ->
+    create_block_ok().
+create_block(Metadata, Txns, HBBFTRound, Chain, N, F, {MyPubKey, SignFun}) ->
+    {ok, CurrentBlock} = blockchain:head_block(Chain),
+    Height_Curr = blockchain_block:height(CurrentBlock),
```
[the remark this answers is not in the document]
xandkar: Fixed

src/miner_util.erl
xandkar: True, but this way the sort order is not obscured!

src/miner.erl
```diff
@@ -600,9 +445,234 @@ terminate(Reason, _State) ->
...
+    blockchain:blockchain(),
+    swarm_keys()
+) ->
+    create_block_result().
+try_create_block(Metadata, Txns, HBBFTRound, Chain, SwarmKeys) ->
+    %% This can actually be a stale message, in which case we'd produce a block
+    %% with a garbage timestamp. This is not actually that big of a deal, since
+    %% it won't be accepted, but we can short circuit some effort by checking
+    %% for a stale hash.
+    Chain = blockchain_worker:blockchain(),
```
evanmcc, line 465: why are we refetching the chain?

src/miner.erl
```diff
@@ -600,9 +445,234 @@ terminate(Reason, _State) ->
...
+    lager:debug("Metadata ~p, current hash ~p", [Metadata, HashCurr]),
+    N = count_consensus_members(Chain),
+    VotesNeeded = N - ((N - 1) div 3),
+    {_, Hashes} = meta_to_stamp_hashes(Metadata),
+    case hash_check_if_stale(HashCurr, Hashes, VotesNeeded) of
+        {ok, {}} ->
+            {ok, create_block(Metadata, Txns, HBBFTRound, Chain, VotesNeeded, SwarmKeys)};
+        {error, {stale, HashStale, Votes}} ->
+            %% TODO What is the appropriate log level here?
+            lager:info("Stale hash: ~p, picked by ~b nodes.", [HashStale, Votes]),
```
evanmcc, line 476: I think go with `warning`

src/miner.erl
xandkar: Great catch! Artifact of moving things around...

xandkar: Fixed

src/miner.erl
```diff
@@ -600,9 +445,232 @@ terminate(Reason, _State) ->
...
+        unsigned_binary_block => BinNewBlock,
+        signature             => Signature,
+
+        %% Both pending and invalid are to be removed from the buffer:
+        pending_txns          => TxnsToInsert,
+        invalid_txns          => InvalidTransactions
+    }.
+
+-spec block_time(blockchain_block:block(), metadata()) -> pos_integer().
+block_time(Block, Metadata) ->
```
evanmcc, line 559: `block_time` -> `median_metadata_timestamp`

src/miner.erl
```diff
@@ -600,9 +445,232 @@ terminate(Reason, _State) ->
...
+        end.
+
+-spec txn_is_rewards(blockchain_txn:txn()) -> boolean().
+txn_is_rewards(Txn) ->
+    Rewards = [blockchain_txn_rewards_v1, blockchain_txn_rewards_v2],
+    lists:member(blockchain_txn:type(Txn), Rewards).
+
+-spec metadata_to_seen_bbas(metadata()) ->
+    [{{pos_integer(), binary()}, binary()}].
+metadata_to_seen_bbas(M) ->
```
evanmcc, line 627: I'm not sure this passes the 'useful abstraction' test

src/miner.erl
```diff
@@ -600,9 +445,232 @@ terminate(Reason, _State) ->
...
+-spec meta_to_stamp_hashes(metadata()) ->
+    {
+        Stamps :: [integer()],
+        Hashes :: [blockchain_block:hash()]
+    }.
+meta_to_stamp_hashes(Metadata) ->
+    lists:unzip([metadata_as_v1(M) || {_, M} <- Metadata]).
+
+-spec metadata_as_v1(metadata_v1() | metadata_v2()) -> metadata_v1().
+metadata_as_v1(#{head_hash := H, timestamp := S}) -> {S, H}; %% v2 -> v1
```
evanmcc: honestly at this point we can remove all the v1 code.  all that it does is complicate things, and nothing will generate it anymore.

src/miner.erl
```diff
@@ -600,9 +445,229 @@ terminate(Reason, _State) ->
...
+    %% agree on one, just leave it blank, so other nodes can absorb it.
+    case blockchain:config(?snapshot_interval, Ledger) of
+        {ok, Interval} when (BlockHeightNext - 1) rem Interval == 0 ->
+            Hashes = [H || {_, #{snapshot_hash := H}} <- metadata_only_v2(Metadata)],
+            pick_consented_or_default(VotesNeeded, Hashes, <<>>);
+        _ ->
+            <<>>
+    end.
+
+-spec pick_consented_or_default(non_neg_integer(), [X], X) -> X.
```
evanmcc, line 663: name is odd? `most_common_over_threshold` maybe?

src/miner.erl
```diff
@@ -600,9 +445,229 @@ terminate(Reason, _State) ->
...
+
+-spec meta_to_stamp_hashes(metadata()) ->
+    {
+        Stamps :: [integer()],
+        Hashes :: [blockchain_block:hash()]
+    }.
+meta_to_stamp_hashes(Metadata) ->
+    lists:unzip([metadata_as_v1(M) || {_, M} <- Metadata]).
+
+-spec metadata_as_v1(metadata_v1() | metadata_v2()) -> metadata_v1().
```
evanmcc, line 643: I think that we can remove all the v1 stuff?  it was for backwards compatibility, and I think that we can toss it now.

src/miner.erl
```diff
@@ -600,9 +445,229 @@ terminate(Reason, _State) ->
...
+    C :: non_neg_integer().
+hash_check_if_stale(HashCurr, Hashes, VotesNeeded) ->
+    case
+        lists:filter(
+            fun ({_, Votes}) -> Votes >= VotesNeeded end,
+            maps:to_list(miner_util:list_count(Hashes))
+        )
+    of
+        %% We expect every stamp to contain the same block hash:
+        [{HashCurr, _}]      -> {ok, {}};
```
evanmcc, line 504: can this just be `ok`?  I don't understand the purpose of the empty tuple here.

evanmcc (commented):
a few things to clean up, a few nits.  looks like it's moving in the right direction, though.

src/miner.erl
xandkar: It was one of the first things to get plucked out of create_block as I was desperately trying to slim it down. It itself wasn't slim either... After several passes - it now looks this slim and can probably be put back in.

src/miner.erl
xandkar: Initially, Professor said that too, but then backtracked - I forgot exactly the reason, but it had something to do with metadata version coming from blockchain config: `blockchain:config(?election_version, Ledger)` https://github.com/helium/miner/blob/b31adb0f93a190fdd2d9e5a744b731407501df01/src/handlers/miner_hbbft_handler.erl#L492

src/miner.erl
xandkar: `most_common_or_default`?

xandkar: `common_enough_or_default`?

Reason why I'm stuck on picking `default` over `threshold` in the name: threshold seems reasoably-obvious in relation to commonality, whereas falling back on default is not.

src/miner.erl
xandkar: That doesn't seem quite right to me: `block_time` describes WHAT the function computes, whereas `median_metadata_timestamp` describes a subset of of HOW it computes it.

Perhaps `new_block_time` or `next_block_time` or something like that? (If you object because `block_time` sounds like an accessor or something...)

src/miner.erl
xandkar: Short answer: yes, it can - no probs, but...

Longer answer: it is a bigger issue than it seems and the solution has to break either intuition or composability and semantic consistency - I chose to break intuition.

Why `{ok, A}` vs `ok`? Compatibility with the general result type `{ok, A} | {error, B}` - so that a function is composable with others that return a Result.

If you accept the virtue of the goal of composability, then you have to choose an `A` that represent nothingness/void. Theoretically it is a unit type and is represented as a 0-tuple in theory, SML, OCaml, Haskell and Rust: https://en.wikipedia.org/wiki/Unit_type

We _could_ also choose some atom, like `ok`, `unit`, `void`, etc., but I find these options even more arbitrary and less satisfying.

src/miner.erl
xandkar: Removed.

src/miner.erl
evanmcc: next block time is fine

evanmcc (approved):

PR #718, opened 2021-04-01
add touchfile follow mode

evanmcc:
Make it so we can put a miner into follow_only mode by touching a file on its drive.  This makes the miner not challenge, so that low-validation miners do not enter into the consensus group.


src/miner_app.erl
```diff
@@ -28,6 +28,15 @@ start(_StartType, _StartArgs) ->
                       BBOpts ++ [{block_cache, Cache}]}],
     application:set_env(rocksdb, global_opts, Opts),
 
+    BaseDir = application:get_env(blockchain, base_dir, "data"),
+    Filename = filename:join([BaseDir, "follow-mode"]),
```
vihu, line 32: wouldn't this allow a directory named "follow-mode" as well?

vihu (approved):

PR #722, opened 2021-04-07
add erlfmt, run on all files

evanmcc:
open for comments

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,39 +1,47 @@
 %% -*- erlang -*-
 {erl_opts, [
-            debug_info,
-            {parse_transform, lager_transform},
-            {i, "./_build/default/plugins/gpb/include"}
-           ]}.
+    debug_info,
+    {parse_transform, lager_transform},
+    {i, "./_build/default/plugins/gpb/include"}
+]}.
 
 {cover_export_enabled, true}.
 {cover_enabled, true}.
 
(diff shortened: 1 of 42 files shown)


rebar.config
```diff
@@ -1,39 +1,47 @@
...
+
+{project_plugins, [
+    erlfmt,
+    {rebar3_eqc, "1.3.0"}
+]}.
+
+{erlfmt, [
+    %% verbose,
+    write,
+    {print_width, 100}
```
JayKickliter, line 39: I'm not a fan of specifying this. If you want to format a single file with the `escript` version of `erlfmt`, e.g. via an editor command, this setting doesn't get picked up and the reformatted results may differ from what `rebar3 fmt` would do.

JayKickliter: (I admit this is a ridiculous state of affairs compared to say `clang-format` with its `some-repo/.clang-format` file)

evanmcc: I guess my hope here was to have this be in a pre-commit hook or something like that, so we wouldn't have to specify it.  `rebar3 fmt` takes less than a second on my machine though, so maybe a single-file version isn't needed?

Vagabond: I believe rebar3 fmt can also be told to crash if the file is not formatted, so we could make that part of CI

PR #733, opened 2021-04-14
switch to threshold crypto

evanmcc:
rebased to just the minimum.


Vagabond:
We can close this, right?

PR #744, opened 2021-04-26
Add an MVP implementation of miner-side discovery mode

JayKickliter:
This PR adds an MVP implementation for miner-side discovery mode.

Outstanding questions:

- what to do when router requests miner to enter a period of discovery mode but the previous batch of uplinks is not complete?
- what to do when `miner_discovery_worker` dies?


macpie:
> what to do when router requests miner to enter a period of discovery mode but the previous batch of uplinks is not complete?

I would think we should then ignore it to avoid spam

> what to do when `miner_discovery_worker` dies?

If you mean normally I would do nothing, its job is done.

macpie:
For the ignore, as your process is named a simple `is_process_alive` should tell you if it is still running.

src/handlers/miner_discovery_handler.erl
```diff
@@ -0,0 +1,123 @@
...
+            lager:info("failed to verify signature for ~p (sig=~p)", [
+                blockchain_utils:addr2name(HostpostPubKeyBin),
+                Sig
+            ]);
+        true ->
+            case miner_lora:location_ok() of
+                true ->
+                    miner_discovery_worker:start(Packets);
+                false ->
+                    lager:warning("can not start discovery mode until location is ok"),
```
michaeldjeffrey, line 77: Do Miner's have a way of sending error messages back? So we don't have to burrow to see if something might not be happening?

src/miner.erl
```diff
@@ -310,6 +310,8 @@ version() ->
 init(_Args) ->
     lager:info("STARTING UP MINER"),
     ok = blockchain_event:add_handler(self()),
+    %% TODO: Maybe put this somewhere else?
+    ok = miner_discovery_handler:add_stream_handler(blockchain_swarm:tid()),
```
Vagabond, line 314: Add it to the supervision tree

src/miner_discovery_worker.erl
```diff
@@ -0,0 +1,118 @@
...
+    timer:send_after(?DEFAULT_TRANSMIT_DELAY_MS, self(), tick),
+    {noreply, State#state{packets = Packets}};
+handle_info(_Info, State) ->
+    {noreply, State}.
+
+%% ------------------------------------------------------------------
+%% Internal
+%% ------------------------------------------------------------------
+
+%% TODO: taken from miner_onion_server, refactor to common module?
```
vihu, line 94: It's fine for now, we should be able to lookup tx_power and spreading using the location of the hotspot once some of the PoC related work lands.

src/handlers/miner_discovery_handler.erl
```diff
@@ -0,0 +1,123 @@
...
+handle_info(_Type, _Msg, State) ->
+    lager:warning("test ~p got info ~p", [_Type, _Msg]),
+    {noreply, State}.
+
+%% ------------------------------------------------------------------
+%% Internal Function Definitions
+%% ------------------------------------------------------------------
+
+-spec verify_signature(
+    Hostpost :: binary(),
```
vihu, line 98: Hostpost -> Hotspot
HostpostPubkeyBin -> HotspotPubkeyBin

:)

src/handlers/miner_discovery_handler.erl
```diff
@@ -0,0 +1,123 @@
...
+    {ok, #state{}}.
+
+handle_data(server, Data, State) ->
+    lager:info("got ~p ~p", [Data, State]),
+    #discovery_start_pb{
+        hotspot = HotspotPubkeyBin,
+        packets = Packets,
+        signature = Sig
+    } = discovery_pb:decode_msg(Data, discovery_start_pb),
+    Hotspot = erlang:list_to_binary(libp2p_crypto:bin_to_b58(HotspotPubkeyBin)),
```
vihu, line 65: Seems a little odd that we're converting a binary to a binary again? If there's a particular reason to do so, maybe add a comment here?

evanmcc (approved):

PR #757, opened 2021-05-04
switch to threshold crypto

Vagabond:


evanmcc (approved):
re-tested the transition with run.sh from 7f345a3ee3dbf3cafac4651c5d6c582921203afd to this branch, everything worked fine.

PR #784, opened 2021-05-23
Use the [[ operator not [

mrallen1:
Problem: `[` only works with numbers

### DIFF
--- a/.buildkite/scripts/make_image.sh
+++ b/.buildkite/scripts/make_image.sh
@@ -16,7 +16,7 @@ docker build -t helium:$DOCKER_NAME -f "${DOCKERFILE_NAME}" .
 docker tag helium:$DOCKER_NAME "$MINER_REGISTRY_NAME:$DOCKER_NAME"
 docker push "$MINER_REGISTRY_NAME:$DOCKER_NAME"
 # only update latest if github tag contains `_GA$`
-if [ "$BUILDKITE_TAG" =~ _GA$ ]; then
+if [[ "$BUILDKITE_TAG" =~ _GA$ ]]; then
     docker tag helium:$DOCKER_NAME "$MINER_REGISTRY_NAME:latest-${IMAGE_ARCH}"
     docker push "$MINER_REGISTRY_NAME:latest-${IMAGE_ARCH}"
 fi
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/make_image.sh
```diff
@@ -16,7 +16,7 @@ docker build -t helium:$DOCKER_NAME -f "${DOCKERFILE_NAME}" .
 docker tag helium:$DOCKER_NAME "$MINER_REGISTRY_NAME:$DOCKER_NAME"
 docker push "$MINER_REGISTRY_NAME:$DOCKER_NAME"
 # only update latest if github tag contains `_GA$`
-if [ "$BUILDKITE_TAG" =~ _GA$ ]; then
+if [[ "$BUILDKITE_TAG" =~ _GA$ ]]; then
```
JayKickliter, line 19: I had to look up this syntax. The trailing $ confused me until I saw that =~ was bash’s regex operator (so we’re looking for _GA at end of the line). 

I’ll approve this, but I almost think grep call makes sense here

JayKickliter (approved):

.buildkite/scripts/make_image.sh
JayKickliter: Actually, I’m even more confused now. The comment above implies that we’re searching for a literal dollar sign. Is that correct?

mrallen1: I can add additional comment if you feel it would be more helpful

JayKickliter: Yes please. Or at least correct the comment if we are not looking for a literal '$' but using it in the regex sense for end of line

.buildkite/scripts/make_image.sh
```diff
@@ -16,7 +16,7 @@ docker build -t helium:$DOCKER_NAME -f "${DOCKERFILE_NAME}" .
 docker tag helium:$DOCKER_NAME "$MINER_REGISTRY_NAME:$DOCKER_NAME"
 docker push "$MINER_REGISTRY_NAME:$DOCKER_NAME"
 # only update latest if github tag contains `_GA$`
```
abhay, line 18:
```suggestion
# only update latest if github tag contains `_GA`
```

vihu (approved):
lgtm

PR #799, opened 2021-05-31
add time-based skip voting

evanmcc:
we never time out a round, which means that if some critical messages are lost, we might never proceed.  This code adds time-based voting to skip.

todo: needs to be properly set on restore?

### DIFF
--- a/src/handlers/miner_hbbft_handler.erl
+++ b/src/handlers/miner_hbbft_handler.erl
@@ -37,7 +37,8 @@
          bba = <<>> :: binary(),
 
          %% For artifact re-signing on var-autoskip:
-         swarm_keys :: {libp2p_crypto:pubkey(), libp2p_crypto:sig_fun()}
+         swarm_keys :: {libp2p_crypto:pubkey(), libp2p_crypto:sig_fun()},
+         skip_votes = #{} :: #{pos_integer() => pos_integer()}
         }).
 
 -type hbbft_msg() ::
@@ -235,6 +236,14 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
                     {reply, ok, fixup_msgs(Msgs), State#state{hbbft=NewHBBFT}}
             end
     end;
+handle_command(maybe_skip, State = #state{hbbft = HBBFT,
+                                          id = MyIndex,
+                                          skip_votes = Skips}) ->
+    MyRound = hbbft:round(HBBFT),
+    %% put in a fake local vote here for the case where we have no skips
+    ProposedRound = median_not_taken(maps:merge(#{MyIndex => {0, MyRound}}, Skips)),
+    lager:info("voting for round ~p", [ProposedRound]),
+    {reply, ok, [{multicast, term_to_binary({proposed_skip, ProposedRound, MyRound})}], State};
 handle_command(_, _State) ->
     {reply, ignored, ignore}.
 
@@ -243,7 +252,7 @@ handle_command(_, _State) ->
     RelcastMsg ::
           {multicast, binary()}
         | {unicast, pos_integer(), binary()}.
-handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft=HBBFT}=S0) ->
+handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft = HBBFT, skip_votes = Skips}=S0) ->
     CurRound = hbbft:round(HBBFT),
     case bin_to_msg(BinMsgIn) of
         %% Multiple sigs
@@ -262,6 +271,39 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft=HBBFT}=S0) ->
         {ok, {signature, _, Addr, Sig}} ->
             handle_sigs({received, [{Addr, Sig}]}, S0);
 
+        {ok, {proposed_skip, ProposedRound, IndexRound}} when ProposedRound =/= CurRound ->
+            lager:info("proposed skip to round ~p", [ProposedRound]),
+            case process_skips(ProposedRound, IndexRound, S0#state.f, Index, Skips) of
+                {skip, Skips1} ->
+                    lager:info("skipping"),
+                    %% ask for some time
+                    miner:reset_late_block_timer(),
+                    SkipGossip = {multicast, t2b({proposed_skip, ProposedRound, CurRound})},
+                    %% skip but don't discard rounds until we get a clean round
+                    {HBBFT1, ToSend1} =
+                        case hbbft:next_round(HBBFT, ProposedRound, []) of
+                            {H1, ok} ->
+                                {H1, []};
+                            {H1, {send, NextMsgs}} ->
+                                {H1, NextMsgs}
+                        end,
+                    {HBBFT2, ToSend2} =
+                        case hbbft:start_on_demand(HBBFT1) of
+                            {H2, already_started} ->
+                                {H2, []};
+                            {H2, {send, Msgs}} ->
+                                {H2, Msgs}
+                        end,
+                    ToSend = fixup_msgs(ToSend1 ++ ToSend2),
+                    {(state_reset(HBBFT2, S0))#state{skip_votes = Skips1},
+                     [ new_epoch , SkipGossip ] ++ ToSend};
+                {wait, Skips1} ->
+                    lager:info("waiting: ~p", [Skips1]),
+                    {S0#state{skip_votes = Skips1}, []}
+            end;
+        {proposed_skip, ProposedRound, IndexRound} ->
+            Skips1 = Skips#{Index => {ProposedRound, IndexRound}},
+            {S0#state{skip_votes = Skips1}, []};
         %% Other
         {ok, MsgIn} ->
             handle_msg_hbbft(MsgIn, Index, S0);
@@ -371,6 +413,8 @@ serialize(State) ->
                         M#{K => term_to_binary(SerializedSK,  [compressed])};
                    ({chain, _}, M) ->
                         M;
+                   ({skip_votes, _}, M) ->
+                        M;
                    ({K, V}, M)->
                         VB = term_to_binary(V, [compressed]),
                         M#{K => VB}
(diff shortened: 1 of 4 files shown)


src/handlers/miner_hbbft_handler.erl
```diff
@@ -529,5 +577,42 @@ md_version(Ledger) ->
             tuple
     end.
 
+%% do nothing if we've advanced
+process_skips(Proposed, SenderRound, F, Sender, Votes) ->
```
vihu, line 581:
```suggestion
-spec process_skips(Proposed :: non_neg_integer(), SenderRound :: non_neg_integer(), F :: non_neg_integer(), Sender :: non_neg_integer(), Votes ::map()) -> {skip | wait, map()}.
process_skips(Proposed, SenderRound, F, Sender, Votes) ->
```

src/handlers/miner_hbbft_handler.erl
```diff
@@ -529,5 +577,42 @@ md_version(Ledger) ->
...
+%% the idea here is to take a clean round that's higher than the median, which should be relatively
+%% hard to manipulate by cheating
+median_not_taken(Map) ->
+    {_Votes, Rounds} = lists:unzip(maps:values(Map)),
+    lager:info("rounds ~p", [lists:sort(Rounds)]),
+    Median = miner_util:median(Rounds),
+    search_lowest(Median, lists:sort(Rounds)).
+
+%% if we haven't found a round in the list, the next highest should work
+search_lowest(Try, []) ->
```
vihu, line 605: Can we add a simple eunit for this particular function?

src/miner.erl
```diff
@@ -347,6 +354,13 @@ handle_call({create_block, Metadata, Txns, HBBFTRound}, _From,
     {reply, Result, State};
 handle_call(keys, _From, State) ->
     {reply, {ok, State#state.swarm_keys}, State};
+handle_call(reset_late_block_timer, _From, State) ->
+    %% we do this when the group thinks that it's agreed on a new round, so set the timer extra long
+    erlang:cancel_timer(State#state.late_block_timer),
+    LateBlockTimeout = application:get_env(miner, late_block_timeout_seconds, 120),
+    LateTimer = erlang:send_after((LateBlockTimeout * 2) * 1000, self(), late_block_timeout),
```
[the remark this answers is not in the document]
vihu: I guess this is fine since it's an app env?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -240,9 +250,43 @@ handle_message(BinMsg, Index, State) when is_binary(BinMsg) ->
...
+                        case hbbft:start_on_demand(HBBFT1) of
+                            {H2, already_started} ->
+                                {H2, []};
+                            {H2, {send, Msgs}} ->
+                                {H2, Msgs}
+                        end,
+                    ToSend = fixup_msgs(ToSend1 ++ ToSend2),
+                    {State#state{hbbft=HBBFT2, signatures=[], artifact=undefined, sig_phase=unsent,
+                                 bba = <<>>, seen = #{}, skip_votes = Skips1},
+                     [ new_epoch , SkipGossip ] ++ ToSend};
```
vihu, line 282: Is this not supposed to be tagged multicast here for `{new_epoch, SkipGossip}`?

test/miner_blockchain_SUITE.erl
```diff
@@ -170,6 +170,58 @@ init_per_testcase(TestCase, Config0) ->
...
+    Node1 = hd(MinersCGBroken),
+    ok = ct_rpc:call(Node1, miner, hbbft_skip, [], 300),
+    ok = ct_rpc:call(Node1, miner, hbbft_skip, [], 300),
+    ok = ct_rpc:call(Node1, miner, hbbft_skip, [], 300),
+    ok = ct_rpc:call(Node1, miner, hbbft_skip, [], 300),
+    ok = ct_rpc:call(Node1, miner, hbbft_skip, [], 300),
+    ok = ct_rpc:call(Node1, miner, hbbft_skip, [], 300),
+
+    ok = miner_ct_utils:assert_chain_halted(MinersAll),
+    ok = miner_ct_utils:assert_chain_advanced(MinersAll, 2500, 5),
```
xandkar, line 222: I'm not following what's going on here. `assert_chain_halted` checks every 5 seconds for over a minute that height is stuck on all nodes and after all that - `assert_chain_advanced` also passes? Is it due to getting the timing right with just the right number of `hbbft_skip` calls?

xandkar: Also, why then focus on one of the nodes?

The test could use a comment explaining the idea behind this magic number of `hbbft_skip` calls.

src/handlers/miner_hbbft_handler.erl
xandkar, line 282: Lack of indentation makes it hard to tell what is what at a glance.

```suggestion
                    ToSend = [new_epoch , SkipGossip] ++ fixup_msgs(ToSend1 ++ ToSend2),
                    State1 = State#state{
                        hbbft = HBBFT2,
                        signatures = [],
                        artifact = undefined,
                        sig_phase = unsent,
                        bba = <<>>,
                        seen = #{},
                        skip_votes = Skips1
                    },
                    {State1, ToSend};
```

src/handlers/miner_hbbft_handler.erl
xandkar: Also a spec.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -529,5 +577,42 @@ md_version(Ledger) ->
...
+                          Votes1),
+    %% equal here because we only want to go once, at the point of saturation
+    case PropVotes == (2 * F) + 1 of
+        true -> {skip, Votes1};
+        false -> {wait, Votes1}
+    end.
+
+%% the idea here is to take a clean round that's higher than the median, which should be relatively
+%% hard to manipulate by cheating
+median_not_taken(Map) ->
```
xandkar, line 598: spec

src/handlers/miner_hbbft_handler.erl
Vagabond: no, new epoch is an command to relcast, not a message

src/handlers/miner_hbbft_handler.erl
```diff
@@ -229,6 +231,14 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
                     {reply, ok, fixup_msgs(Msgs), State#state{hbbft=NewHBBFT}}
             end
     end;
+handle_command(maybe_skip, State = #state{hbbft = HBBFT,
```
Vagabond, line 234: I would argue that this should replace the old skip command entirely, as the old one is no longer safe and very easily leads to misalignment

src/miner.erl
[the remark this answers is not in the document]
Vagabond: it's double the normal timer since we just agreed on a skip

Vagabond (commented):
This overall looks ok, but I would echo the existing review comments and ask for a bit more explanation of the rationale about median_not_taken and how it helps

src/handlers/miner_hbbft_handler.erl
evanmcc: done

src/handlers/miner_hbbft_handler.erl
evanmcc: done

test/miner_blockchain_SUITE.erl
evanmcc: added some comments.   the double check is just to make sure that the chain halts for a while and then goes forward again.  I guess there might be a quicker way to figure that out?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -262,6 +260,39 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft=HBBFT}=S0) ->
...
+                        end,
+                    {HBBFT2, ToSend2} =
+                        case hbbft:start_on_demand(HBBFT1) of
+                            {H2, already_started} ->
+                                {H2, []};
+                            {H2, {send, Msgs}} ->
+                                {H2, Msgs}
+                        end,
+                    ToSend = fixup_msgs(ToSend1 ++ ToSend2),
+                    {(state_reset(HBBFT2, S0))#state{skip_votes = Skips1},
```
Vagabond, line 287: Why do we retain the skip votes here?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -698,6 +734,49 @@ kvl_set(K, V, KVL) ->
 b58(<<Bin/binary>>) ->
     libp2p_crypto:bin_to_b58(Bin).
 
+%% do nothing if we've advanced
+process_skips(Proposed, SenderRound, F, Sender, Votes) ->
+    Votes1 = Votes#{Sender => {Proposed, SenderRound}},
```
Vagabond, line 739: This ensures that each actor can only have one vote at a time, right? Are there protections against an actor changing its vote, or sending different votes to different peers or is this not a risk?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -262,6 +260,39 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft=HBBFT}=S0) ->
...
 
+        {ok, {proposed_skip, ProposedRound, IndexRound}} when ProposedRound =/= CurRound ->
+            lager:info("proposed skip to round ~p", [ProposedRound]),
+            case process_skips(ProposedRound, IndexRound, S0#state.f, Index, Skips) of
+                {skip, Skips1} ->
+                    lager:info("skipping"),
+                    %% ask for some time
+                    miner:reset_late_block_timer(),
+                    SkipGossip = {multicast, t2b({proposed_skip, ProposedRound, CurRound})},
+                    %% skip but don't discard rounds until we get a clean round
```
Vagabond, line 271: How are avoiding the discard of previous rounds?

src/handlers/miner_hbbft_handler.erl
evanmcc: we retain them until we complete a block, in case we fail and new members join and we converge on a different round.

src/handlers/miner_hbbft_handler.erl
Vagabond: Change 'rounds' to 'votes' here

Vagabond (approved):

PR #812, opened 2021-06-08
penalty txns are invalid before election v5, so don't start them

evanmcc:
this is one of the last blockers before we merge the `pevm/validators` branch to master.  ideally we would not waste resources on these before we need to.


Vagabond:
Should this not use the pinned ledger from `{add_block, Hash, Sync, _Ledger}`?

Vagabond (approved):

PR #813, opened 2021-06-09
Don't participate in consensus if follow mode is enabled

Vagabond:


src/miner_consensus_mgr.erl
```diff
@@ -548,15 +553,20 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
             {noreply, State}
     end;
 handle_info({blockchain_event, {add_block, _Hash, _Sync, _Ledger}}, State) ->
-    case State#state.chain of
-        undefined ->
+    case {State#state.chain, application:get_env(blockchain, follow_mode, false)} of
+        {undefined, false} ->
```
andymck, line 557: do you need to handle {undefined, true} here ?

PR #815, opened 2021-06-09
Test time-based skip voting

xandkar:
TODO
- [x] branch from `pevm/validators` and reproduce chain stall
- [x] rebase to `pevm/time-based-skip` and switch test to expect chain advance instead of stall


xandkar:
> it should be easy enough to halt the chain by using rpc to call miner:hbbft_skip on some consensus members.

lol - prof explicitly told me _not_ to use `miner:hbbft_skip/0`

PR #802, opened 2021-06-02
Rework miner_ct_utils:initial_dkg to account for partial completion

Vagabond:

PR #820, opened 2021-06-11
refactor txn mgr dependency test

mrallen1:
Rework miner_ct_utils:initial_dkg to account for partial completion

Refactor tests for new miner_ct_utils:initial_dkg

This is to land #802 retargeted to the master branch


rebar.config
```diff
@@ -10,7 +10,7 @@
 
 {deps, [
         {blockchain, {git, "https://github.com/helium/blockchain-core.git",
-                      {branch, "master"}}},
+                      {branch, "andymck/check-nonce-during-isvalid"}}},
```
evanmcc, line 13: let's pull this from the master version

evanmcc (approved):
get rid of the config change and I think that this is ready to merge

rebar.config
mrallen1: ok, updated to rebar.config from master

PR #822, opened 2021-06-14
Prep HBBFT handler for adding var autoskip

xandkar:


src/handlers/miner_hbbft_handler.erl
```diff
@@ -21,22 +21,43 @@
         {
          n :: non_neg_integer(),
          f :: non_neg_integer(),
-         id :: non_neg_integer(),
+         id :: non_neg_integer(), % TODO Remove? Set at init, but never accessed.
```
Vagabond, line 24: This is used in #799

src/handlers/miner_hbbft_handler.erl
```diff
@@ -21,22 +21,43 @@
...
+         deferred = [],  % TODO Remove? Looks unused.
+         sigs_valid     = [] :: sigs(),
+         sigs_invalid   = [] :: sigs(),
+         sigs_unchecked = [] :: sigs(),
          signatures_required = 0,
          sig_phase = unsent :: unsent | sig | gossip | done,
          artifact :: undefined | binary(),
          members :: [libp2p_crypto:pubkey_bin()],
-         chain :: undefined | blockchain:blockchain(),
+         chain :: undefined | blockchain:blockchain(), % TODO Remove? Set at init, but never accessed.
```
Vagabond, line 36: Remove if not used, yes

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
-    false;
-enough_signatures(#state{sig_phase = done, signatures = Signatures}) ->
-    {ok, done, Signatures};
-enough_signatures(#state{sig_phase = Phase, artifact = Artifact, members = Members,
-                         f = F, signatures = Signatures, signatures_required = Threshold0}) ->
-    Threshold = case Phase of
-                    sig -> F + 1;
-                    gossip -> Threshold0
-                end,
+-spec sigs_to_positions([{A, _, _}], [A]) -> [pos_integer()].
```
Vagabond, line 486:
```suggestion
% @doc Converts a list of `{Address, Signature}' tuples into a a list of the Address' position in the `Addresses' list.
-spec sigs_to_positions([{Address :: binary(), Signature :: }], Addresses :: [binary()]) -> [pos_integer()].
```

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
+    Addr2Pos = lists:zip(Addresses, lists:seq(1, length(Addresses))),
+    Positions = lists:map(fun({Addr, _S, _V}) ->
+                            %% find member index
+                            {_, Pos} = lists:keyfind(Addr, 1, Addr2Pos),
+                            Pos
+                    end,
+                    Addr2Sig),
+    lists:usort(Positions).
+
+md_version(Ledger) ->
```
Vagabond, line 497:
```suggestion
-spec md_version(Ledger :: blockchain_ledger_v1:ledger()) -> map | tuple.
md_version(Ledger) ->
```

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
+    CurRound = hbbft:round(HBBFT),
+    S1 = state_sigs_add(SigsIn, S0),
+    S2 = state_sigs_retry_unchecked(S1),
+    case state_sigs_check_if_enough(S2) of
+        {{some, {gossip, _}=Gossip}, S3} ->
+            {Gossip, S3#state{sig_phase = gossip}};
+        %% XXX What if artifact is undefined?
+        {{some, {done, Sigs}=Done}, #state{artifact=Art, signed=Signed}=S3} when CurRound > Signed ->
+            %% no point in doing this more than once
+            % TODO ^^^ Doing what? miner:signed_block or ?mark(done_sigs) ?
```
Vagabond, line 533: Signed block

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
+        #state{}
+    }.
+state_sigs_input(SigsIn, #state{hbbft=HBBFT}=S0) ->
+    CurRound = hbbft:round(HBBFT),
+    S1 = state_sigs_add(SigsIn, S0),
+    S2 = state_sigs_retry_unchecked(S1),
+    case state_sigs_check_if_enough(S2) of
+        {{some, {gossip, _}=Gossip}, S3} ->
+            {Gossip, S3#state{sig_phase = gossip}};
+        %% XXX What if artifact is undefined?
```
Vagabond, line 530: You can't check the signatures without an artifact?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
+        {{some, {done, Sigs}=Done}, #state{artifact=Art, signed=Signed}=S3} when CurRound > Signed ->
+            %% no point in doing this more than once
+            % TODO ^^^ Doing what? miner:signed_block or ?mark(done_sigs) ?
+            ok = miner:signed_block(Sigs, Art),
+            S4 =
+                S3#state{
+                    % TODO Q: Why does "signed = CurRound"? Seems like diff terms.
+                    signed = CurRound,
```
Vagabond, line 538: Probably should be renamed 'last_round_signed' or something

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
+                    sig_phase = done
+                },
+            {Done, S4};
+        {{some, {done, _}}, S3} ->
+            {pending, S3};
+        {none, S3} ->
+            {pending, S3}
+    end.
+
+-spec state_sigs_add([{binary(), binary()}], #state{}) -> #state{}.
```
Vagabond, line 548:
```suggestion
-spec state_sigs_add(sigs(), #state{}) -> #state{}.
```

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +478,189 @@ restore(OldState, NewState) ->
...
-                         libp2p_crypto:verify(Artifact, Signature, libp2p_crypto:bin_to_pubkey(Address))
-                 end, Signatures),
-    State#state{signatures=FilteredSignatures}.
+sig_is_valid(
+    {<<Addr/binary>>, <<Sig/binary>>},
+    #state{artifact=Art, members=Mems, hbbft=HBBFT}
+) ->
+    IsVerified = libp2p_crypto:verify(Art, Sig, libp2p_crypto:bin_to_pubkey(Addr)),
+    IsMember = lists:member(Addr, Mems),
+    IsValid = IsMember andalso IsVerified,
```
Vagabond, line 632: I'd recommend doing something like:

`IsValid = lists:member(Addr, Mems) andalso libp2p_crypto:verify(Art, Sig, libp2p_crypto:bin_to_pubkey(Addr)),`

This way we don't bother trying to do an expensive signature check if the address is not even one we know about.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -232,141 +245,149 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
...
+        %% Sing sig
+        {ok, {signature, SigRound, _, _}} when SigRound > CurRound ->
+            defer;
+        {ok, {signature, SigRound, _, _}} when SigRound < CurRound ->
+            ignore;
+        {ok, {signature, _, Addr, Sig}} ->
+            %% TODO Unify the 3 handlings of state_sigs_input output.
+            %% There 2 differences:
+            %% - call to ?mark(gossip_sigs) - 2 have it, 1 doesn't
+            %% - send signature on pending - 1 has it, 2 don't
```
Vagabond, line 283: ?mark(gossip_sigs) is fine to unify.

For sending the signature, we don't have it in all cases, so it cannot be unfied. You could pass in a signature | undefined for the different call sites though.

Vagabond: Specifically we are sending our *own* signature, which we have just obtained over the artifact in that one case.

src/handlers/miner_hbbft_handler.erl
xandkar: Makes sense. I did it this way to avoid recomputing `IsVerified` in the negative logging case below.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +484,189 @@ restore(OldState, NewState) ->
...
+    case IsValid of
+        true ->
+            ok;
+        false ->
+            CurRound = hbbft:round(HBBFT),
+            % TODO Maybe this logging only makes sense in handle_msg_signature?
+            %lager:warning("Invalid signature ~s from ~s for round ~p in our round ~p", [b58(Sig), b58(Addr), SigRound, CurRound]),
+            lager:warning("Invalid signature ~s from ~s in our round ~p", [b58(Sig), b58(Addr), CurRound]),
+            lager:warning("member? ~p", [IsMember]),
+            lager:warning("verified? ~p", [IsVerified])
```
xandkar, line 646: @Vagabond If we remove this `verified?` logging - we can drop recomputing `libp2p_crypto:verify`.

src/handlers/miner_hbbft_handler.erl
xandkar: > ?mark(gossip_sigs) is fine to unify.

It is called after inputting a single sig (i.e. `{signature, ...` msg), but not after inputting multiple sigs (i.e. `{signatures, ...` msg). Do we need to:
- a) preserve as is;
- b) call `?mark(gossip_sigs)` in both places;
- c) remove `?mark(gossip_sigs)` in both places.
?

xandkar: > For sending the signature, we don't have it in all cases, so it cannot be unfied. You could pass in a signature | undefined for the different call sites though.

Makes sense.

Vagabond: > > ?mark(gossip_sigs) is fine to unify.
> 
> It is called after inputting a single sig (i.e. `{signature, ...` msg), but not after inputting multiple sigs (i.e. `{signatures, ...` msg). Do we need to:
> 
>     * a) preserve as is;
> 
>     * b) call `?mark(gossip_sigs)` in both places;
> 
>     * c) remove `?mark(gossip_sigs)` in both places.
>       ?

I think anywhere we transition to the gossip state.

src/handlers/miner_hbbft_handler.erl
Vagabond: I'd prefer to infer it by looking at the membership check

xandkar: Without recomputing we can only go this far:
```erl
Verified =
    case IsMember of
        true -> member_with_invalid_sig;
        false -> non_member_with_unknown_sig_status
    end
```

src/handlers/miner_hbbft_handler.erl
xandkar: Below, `Art` is passed into `ok = miner:signed_block(Sigs, Art),`, which would crash on `blockchain_block:deserialize(BinBlock)` if I'm reading the call chain correctly.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +482,190 @@ restore(OldState, NewState) ->
...
+    case state_sigs_check_if_enough(S2) of
+        {{some, {gossip, _}=Gossip}, S3} ->
+            {Gossip, S3#state{sig_phase = gossip}};
+        {
+            {some, {done, Sigs}=Done},
+            #state{artifact=Art, last_round_signed=LastRoundSigned}=S3
+        } when CurRound > LastRoundSigned ->
+            %% no point in doing this more than once
+            %% XXX What if artifact is undefined?
+            ok = miner:signed_block(Sigs, Art),
```
xandkar, line 543: @Vagabond If this `Art` is `undefined` - `miner:signed_block` will crash on `blockchain_block:deserialize(BinBlock)`, if I'm reading the call chain correctly.

src/handlers/miner_hbbft_handler.erl
xandkar: The couple of times chain is used (in handling `skip` and `next_round`), it is grabbed from `blockchain_worker:blockchain()`, which blocks. Seems like it would be better to start using the chain from state to replace those calls instead of removing. What do you think? @Vagabond @evanmcc

Vagabond: yes, that is also fine

src/handlers/miner_hbbft_handler.erl
```diff
@@ -457,77 +482,190 @@ restore(OldState, NewState) ->
...
+state_sigs_input(SigsIn, #state{hbbft=HBBFT}=S0) ->
+    CurRound = hbbft:round(HBBFT),
+    S1 = state_sigs_add(SigsIn, S0),
+    S2 = state_sigs_retry_unchecked(S1),
+    case state_sigs_check_if_enough(S2) of
+        {{some, {gossip, _}=Gossip}, S3} ->
+            {Gossip, S3#state{sig_phase = gossip}};
+        {
+            {some, {done, Sigs}=Done},
+            #state{artifact=Art, last_round_signed=LastRoundSigned}=S3
```
xandkar, line 539:
```suggestion
            #state{artifact=<<Art/binary>>, last_round_signed=LastRoundSigned}=S3
```

src/handlers/miner_hbbft_handler.erl
```diff
@@ -21,22 +21,49 @@
...
          f :: non_neg_integer(),
-         id :: non_neg_integer(),
+         id :: non_neg_integer(), % TODO Remove? Set at init, but never accessed.
          hbbft :: hbbft:hbbft_data(),
          sk :: tc_key_share:tc_key_share() | tpke_privkey:privkey() | binary(),
-         seq = 0,
-         deferred = [],
-         signatures = [],
+         seq = 0,        % TODO Remove? Looks unused.
+         deferred = [],  % TODO Remove? Looks unused.
```
xandkar, line 28: Yes. Remove.

evanmcc (approved):

PR #826, opened 2021-06-15
Make DKG completion a bit more tolerant

Vagabond:


src/miner_consensus_mgr.erl
```diff
@@ -216,10 +216,15 @@ handle_call({genesis_block_done, BinaryGenesisBlock, Signatures, Members, PrivKe
     miner_hbbft_sidecar:set_group(Group),
     miner:start_chain(Group, Chain),
 
-    #{ {1,0} := DKGGroup} = State#state.current_dkgs,
+    CancelDKGs = case State#state.current_dkgs of
+        #{ {1,0} := DKGGroup} ->
+            #{3 => DKGGroup};
+        _ ->
+            #{}
```
Vagabond, line 223: Maybe this should be the old value of cancel_dkgs?

evanmcc: mutating the old cancel_dkgs is probably safest, but I really need to remove this stuff to clarify the mechanism.

evanmcc (approved):
make the commented change and you're OK to merge

PR #836, opened 2021-06-21
Extend metadata fun to gossip release info

mrallen1:
Problem to solve: We currently do not collect any firmware release information from miners in their gossiped metadata. It would be a useful thing to know, especially when validators come online and we do not have the ability to inspect them.

Solution: Collect the firmware release info (if it exists in the expected spot and format)

### DIFF
--- a/src/miner_util.erl
+++ b/src/miner_util.erl
@@ -16,6 +16,10 @@
          has_valid_local_capability/2
         ]).
 
+%% get the firmware release data from a hotspot
+-define(LSB_FILE, "/etc/lsb_release").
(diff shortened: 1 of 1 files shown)


src/miner_util.erl
```diff
@@ -90,7 +94,14 @@ metadata_fun() ->
         case application:get_env(miner, mode, gateway) of
             validator ->
                 Vsn = element(2, hd(release_handler:which_releases(permanent))),
-                Map#{<<"release_version">> => list_to_binary(Vsn)};
+                FWRelease = case filelib:is_regular(?LSB_FILE) of
```
evanmcc, line 97: this is the code that should replace the gateway section below, this section was fine as-is

evanmcc (approved):

PR #837, opened 2021-06-21
Remove all tpke references

vihu:
Problem: `miner_dkg_penalty_handler` is missing how to handle `tc_key_share:secret_key` [de]ser

Solution: 

- Add `tc_key_share` functions to `miner_dkg_penalty_handler
- Furhter remove all references of `tpke` from miner src, however there's tpke in the lock file which we can remove once we consolidate the ultimate rebar.lock
- Also add `validator_penalty_filter => 10.0` for run.sh support

Test:
- Confirmed run.sh works after stop/start a node in consensus locally


vihu:
Closing in favor of #839

Issue #882, opened 2021-07-01
allow rpc to bind to 0.0.0.0


kobrien:
RPC only binds to localhost, according to here
https://github.com/helium/miner/blob/e45d8187c267ab02e0c446625c2bb7a7e42e2a5b/config/sys.config#L89

I am able to query RPC within the helium docker container itself
```
docker exec -it helium /bin/sh
/opt/miner # curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"info_height"}' http://localhost:4467
{"jsonrpc":"2.0","result":{"epoch":22935,"height":902163},"id":1}
```

but not from another container on the same docker network
```
docker exec -it nginx-proxy /bin/sh

/ # curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"info_height", "params": {}}' http://helium:4467
curl: (7) Failed to connect to helium port 4467: Connection refused
```

Please allow a way to configure RPC to bind to another address. Thanks!

PR #884, opened 2021-07-01
Bind JSONRPC to 0.0.0.0 for docker

PaulVMo:
Fix for #882. If bond to 127.0.0.1, this will only listen within a docker container making it inaccessible.

### DIFF
--- a/config/docker-val.config.src
+++ b/config/docker-val.config.src
@@ -40,7 +40,7 @@
   ]},
  {miner,
   [
-   {jsonrpc_ip, {0,0,0,0}}, %% bind to host for docker containers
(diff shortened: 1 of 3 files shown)


config/docker-val.config.src
```diff
@@ -40,6 +40,7 @@
   ]},
  {miner,
   [
+   {jsonrpc_ip, {0,0,0,0}}, %% bind to host for docker containers
```
madninja, line 43: nice, thank you! Could do do this for the `docker-testval` and `docker` configs too?

PaulVMo: Done. Thanks for pointing that out.

madninja (approved):

PR #887, opened 2021-07-01
add version to jsonrpc

lthiery:


src/miner_jsonrpc_handler.erl
```diff
@@ -58,6 +58,8 @@ handle_rpc_(<<"sc_", _/binary>> = Method, Params) ->
     miner_jsonrpc_sc:handle_rpc(Method, Params);
 handle_rpc_(<<"peer_", _/binary>> = Method, Params) ->
     miner_jsonrpc_peer:handle_rpc(Method, Params);
+handle_rpc_(<<"version", _/binary>> = Method, Params) ->
```
madninja, line 61: Would this not be better hanging of of info_ ?

lthiery: Good point.

lthiery:
@madninja can we land this one now?

src/jsonrpc/miner_jsonrpc_info.erl
```diff
@@ -78,6 +78,9 @@ handle_rpc(<<"info_summary">>, []) ->
         firmware_version => FirmwareVersion,
         gateway_details => GWInfo
     };
+handle_rpc(<<"info_version">>, []) ->
+    #{version => miner:version()};
```
madninja, line 82: On second review.. why not just add it to the info_summary as `miner_version` in

madninja: bump @lthiery

andrewboudreau: I'm not sure why, yet, but the `info_summary` calls take 2-3 seconds to return most of the time. The response for `miner:version()` is milliseconds.  I like it split into it's own call, but can definitely see the benefits of reducing the api surface and putting into into the summary call.  I do want to be able to get the version over rpc though.

madninja (approved):

PR #888, opened 2021-07-02
Fix missing key_info fields

xandkar:
Not sure if:
- `address` type is restrictive-enough - can it be 0? less than 0?
- `undefined` is a possible value for each


madninja:
> Not sure if:
> 
> * `address` type is restrictive-enough - can it be 0? less than 0?

I think `non_neg_integer()` would cover it

> * `undefined` is a possible value for each

No

xandkar:
> > Not sure if:
> > 
> > * `address` type is restrictive-enough - can it be 0? less than 0?
> 
> I think `non_neg_integer()` would cover it
> 
> > * `undefined` is a possible value for each
> 
> No

Thanks. Updated.

madninja (approved):

PR #889, opened 2021-07-02
Fix undefined chain

xandkar:


src/miner.erl
```diff
@@ -750,6 +752,12 @@ set_next_block_timer(State=#state{blockchain=Chain}) ->
 
     State#state{block_timer=Timer, late_block_timer = LateTimer}.
 
+-spec state_ensure_chain(#state{}) -> #state{}.
+state_ensure_chain(#state{blockchain=undefined}=S) ->
+    S#state{blockchain=blockchain_worker:blockchain()};
```
Vagabond, line 757: This can return undefined

xandkar:
Talked to Evan about this yesterday - he suggests that rather than trying to handle this state _in_ the `miner` module, we should fix the order of calls _to_ the `miner` module so that it is never called in a chainless state.

PR #894, opened 2021-07-05
fix #893 - jsonrpc p2p_status jsx encoding error

### LINKED ISSUE #893: JSONRPC - info_p2p_status
### Issue
 - The `info_p2p_status` command on the JSONPRC endpoint returns `Internal error.`

### Expected Behavior
- The `info_p2p_status` command on the JSONRPC should return the peer to peer status information in JSON format.
- Expected results to be similar to `miner info p2p_status` 

|name| result|
|--|--|
|connected| yes|
|dialable| yes|
|nat_type|none|
|height|15104|

### Further Info
 - curl Command
`curl -H 'Content-Type: application/json-rpc' -d '{"jsonrpc":"2.0","id":1,"method":"info_p2p_status"}' http://localhost:4467`

- curl Result
`{"jsonrpc":"2.0","error":{"code":-32603,"message":"Internal error."},"id":null}`

- Logs
```
2021-07-05 01:11:45.447 [info] <0.20435.1>@miner_jsonrpc_handler:handle_rpc:36 Dispatching method <<"info_p2p_status">> with params: []
2021-07-05 01:11:45.449 [error] <0.20435.1> Failed encoding reply as JSON: {[{<<"jsonrpc">>,<<"2.0">>},{<<"result">>,#{"connected" => "no","dialable" => "yes","height" => "15112","nat_type" => "none"}},{<<"id">>,1}]}
```

andrewboudreau:
Fix for #893 . Add an explicit mapping for `p2p_status` results. Old mapping strategy wasn't compatible with `jsx:encode`.  The new mapping pulls parts from the existing `miner:p2p_status` method and builds a record for each part. `Height` is cast to an integer.

### Usage Example
`curl -H 'Content-Type: application/json-rpc' -d '{"jsonrpc":"2.0","id":1,"method":"info_p2p_status"}'  http://127.0.0.1:4467`

```json
{"jsonrpc":"2.0","result":
    {"p2p_status":
        {"connected":"yes","dialable":"yes","height":907654,"nat_type":"none"}},"id":1}
```


andrewboudreau:
originally I had written out something like,
```erlang
format_p2p_status(Status) ->
  lists:map(fun(T) -> {list_to_atom(element(1,T)), list_to_binary(element(2,T))} end, Status).
```

but wasn't totally happy with either of these path needing to know about the shape of `miner:p2p_status` result. Also this doesn't get `Height` into an integer.

src/jsonrpc/miner_jsonrpc_info.erl
```diff
@@ -31,7 +31,15 @@ handle_rpc(<<"info_name">>, []) ->
 handle_rpc(<<"info_block_age">>, []) ->
     #{block_age => miner:block_age()};
 handle_rpc(<<"info_p2p_status">>, []) ->
-    maps:from_list(miner:p2p_status());
+    [{"connected", Connected}, {"dialable", Dialable}, {"nat_type", NatType}, {"height", Height}] = miner:p2p_status(),
```
madninja, line 34: This patch will break if a new element is added or the ordering of the list changes. . Match on the result from `maps:from_list` to avoid changes to p2p_status output. 

Try something like 

```
#{
    "connected" := Connected,
    "dialable" := Dialable,
... 
} = maps:from_list(miner:p2p_status())
```

madninja (changes requested):

madninja:
@andrewboudreau, bump. Were you going to address the requested changes to your PR?

andrewboudreau:
Yeah, sorry for the delay, team helium. I'll get the PR updated tonight after normal working hours CST. 

If you are prepping a build or otherwise push your own changes I won't be heartbroken seeing as you already spoon-fed the solution here. Which I really appreciate the support, guidance, and patience, thank you.

andrewboudreau:
resulting changes:

```bash
~$ curl -H 'Content-Type: application/json-rpc' -d '{"jsonrpc":"2.0","id":1,"method":"info_p2p_status"}'  http://127.0.0.1:4467
{"jsonrpc":"2.0","result":{"connected":"yes","dialable":"yes","height":919024,"nat_type":"none"},"id":1}
```

src/jsonrpc/miner_jsonrpc_info.erl
andrewboudreau: let me know what you think about the most recent change, I built and tested the results against mainnet, seems to be working well, an example response has been included below.

andrewboudreau: I ended up removing the wrapping p2p_status record, it just feels like overhead, and there isn't 100% consistency between the other calls, but I'd prefer no wrapper when using the results.

madninja (approved):

PR #895, opened 2021-07-06
Make skip messages 'portable' such that they can be gossipped

Vagabond:


src/handlers/miner_hbbft_handler.erl
```diff
@@ -297,6 +297,37 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft = HBBFT, skip_votes = Sk
         {ok, {proposed_skip, ProposedRound, IndexRound}} ->
             Skips1 = Skips#{Index => {ProposedRound, IndexRound}},
             {S0#state{skip_votes = Skips1}, []};
+        {ok, {skip, Round, Sigs}} when Round /= CurRound ->
+            case lists:all(fun({SigIndex, Sig}) ->
+                                   libp2p_crypto:verify(t2b({proposed_skip, Round}), Sig, libp2p_crypto:bin_to_pubkey(lists:nth(SigIndex, S0#state.members)))
```
evanmcc, line 302: we should recheck the length exceeds the threshold here too

Vagabond: good call.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -297,6 +297,40 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft = HBBFT, skip_votes = Sk
...
+                        case hbbft:start_on_demand(HBBFT1) of
+                            {H2, already_started} ->
+                                {H2, []};
+                            {H2, {send, Msgs}} ->
+                                {H2, Msgs}
+                        end,
+                    ToSend = fixup_msgs(ToSend1 ++ ToSend2),
+                    %% we retain the skip votes here because we may not succeed and we want the
+                    %% algorithm to converge more quickly in that case
+                    {(state_reset(HBBFT2, S0))#state{skip_votes = S0#state.skip_votes},
```
Vagabond, line 326: Should we use the skip votes we got in this skip message here?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -297,6 +297,40 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft = HBBFT, skip_votes = Sk
         {ok, {proposed_skip, ProposedRound, IndexRound}} ->
             Skips1 = Skips#{Index => {ProposedRound, IndexRound}},
             {S0#state{skip_votes = Skips1}, []};
+        {ok, {skip, Round, Sigs}} when Round > CurRound ->
```
Vagabond, line 300: Round > CurRound is correct, right? we can never skip backwards?

src/handlers/miner_hbbft_handler.erl
Vagabond: Done.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -804,6 +844,9 @@ bin_to_msg(<<Bin/binary>>) ->
         {error, truncated}
     end.
 
+sign_skip({proposed_skip, ProposedRound, MyRound}, #state{swarm_keys={_Key, SigFun}}) ->
+    {proposed_skip, ProposedRound, MyRound, SigFun(t2b({proposed_skip, ProposedRound}))}.
```
Vagabond, line 848: We only sign over the proposed round, not myround, is that ok?

src/handlers/miner_hbbft_handler.erl
Vagabond: Or do we even still need to do this? the skip was confirmed at this point?

evanmcc: I don't think we need to retain votes if we have echoable votes, no

src/handlers/miner_hbbft_handler.erl
evanmcc: we can skip backwards, we just can't skip onto the existing round, or we won't make any progress.

src/handlers/miner_hbbft_handler.erl
```diff
@@ -297,6 +297,40 @@ handle_message(<<BinMsgIn/binary>>, Index, #state{hbbft = HBBFT, skip_votes = Sk
...
+                            {H2, already_started} ->
+                                {H2, []};
+                            {H2, {send, Msgs}} ->
+                                {H2, Msgs}
+                        end,
+                    ToSend = fixup_msgs(ToSend1 ++ ToSend2),
+                    %% we retain the skip votes here because we may not succeed and we want the
+                    %% algorithm to converge more quickly in that case
+                    {(state_reset(HBBFT2, S0))#state{skip_votes = S0#state.skip_votes},
+                     [ new_epoch | ToSend]};
```
Vagabond, line 327: Should we send the skip message to all the nodes who did not sign the skip here?

src/handlers/miner_hbbft_handler.erl
evanmcc: I think that this will allow some minor adversarial behavior but I am not sure if we care?  If you're not being honest about your round and want the group to stop, you don't plan on executing anyway?

evanmcc (approved):

PR #897, opened 2021-07-06
Shuffle fastforward candidate list to minimise ordering penalties

Vagabond:


src/miner_consensus_mgr.erl
```diff
@@ -527,7 +527,7 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
                                         lists:foreach(
                                           fun(Member) ->
                                               spawn(fun() -> blockchain_fastforward_handler:dial(Swarm, State#state.chain, libp2p_crypto:pubkey_bin_to_p2p(Member)) end)
-                                          end, NewConsensusAddrs -- OldConsensusAddrs),
+                                          end, blockchain_utils:shuffle(NewConsensusAddrs -- OldConsensusAddrs)),
```
evanmcc, line 530: is this expensive enough that we want to subset it for large groups?

evanmcc (approved):

src/miner_consensus_mgr.erl
Vagabond: no, not for small lists (and this is just the list of new nodes being added to the group).

PR #906, opened 2021-07-08
put all gateways into non-validating mode

evanmcc:
since we have moved the consensus group to the validator, there's no point in the gateways validating increasingly expensive transactions as they can no longer participate in consensus.


vihu (approved):
lgtm

Issue #908, opened 2021-07-08
JSONRPC - ledger_validators - "name" should be string, not integer list


ke6jjj:
# Summary

The returned data from the `ledger_validators` call contains a list of currently-staked validators and information about them. One of these items is the validator's angry-purple-tiger name. It is being encoded in the result set as a list of integers. It should be a string, instead.

# To Reproduce

`curl -H 'Content-Type: application/json-rpc' -d '{"jsonrpc":"2.0","id":1,"method":"ledger_validators"}' http://localhost:4467 | jq`

## Result

```json
{
  "jsonrpc": "2.0",
  "result": [
    {
     "comment": "items have been omitted for brevity.",
      "name": [
        102,
        97,
        105,
        116,
        104,
        102,
        117,
        108,
        45,
        111,
        114,
        97,
        110,
        103,
        101,
        45,
        109,
        101,
        101,
        114,
        107,
        97,
        116
      ]
    },
```

ke6jjj:
I'll be fixing a whole slew of issues around the JSON/RPC interface and strings.

PR #911, opened 2021-07-08
JSON/RPC string fixes

ke6jjj:
Fix several points in the JSON/RPC interface where items which should be strings in JSON are instead appearing as lists of integers in result sets and input parameters coming from JSON requests are likewise being misinterpreted by the Erlang.

Fixes helium/miner#908, where I first noticed the issue.


abhay:
This all seems reasonable but let's get a confirmation from core devs

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -157,7 +157,5 @@ last_challenge(_Height, undefined) -> <<"undefined">>;
 last_challenge(Height, LC) -> Height - LC.
 
 format_ledger_validator(Addr, Val, Ledger, Height, Verbose) ->
-    maps:from_list([
-        {<<"name">>, blockchain_utils:addr2name(Addr)}
-        | blockchain_ledger_validator_v1:print(Val, Height, Verbose, Ledger)
-    ]).
+    InfoPairs = blockchain_ledger_validator_v1:print(Val, Height, Verbose, Ledger),
+    maps:from_list(InfoPairs ++ [ {name, ?BIN_TO_ANIMAL(Addr)}, {address, ?BIN_TO_B58(Addr)} ]).
```
madninja, line 161: So passing in verbose already sets `validator_address` in the returned json. 

Either always pass verbose as false, or drop the `name` entry here?

ke6jjj: I see. I think that "address" should always be returned and that verbose, if used at all, should cause the name to be returned as well. Currently that is reversed, and frankl, silly.

But there's more work that can be done with this method, actually. I'm noticing that it returns everything as strings, which is silly for a JSON interface since it can return integers and floats, at the very least, natively.

I'm going to remove the :print() can and instead see if I can implement what it does more directly.

ke6jjj:
By the way, I have tested all of these changes. They seem ok.

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -157,5 +157,33 @@ last_challenge(_Height, undefined) -> <<"undefined">>;
...
+    OwnerAddress = blockchain_ledger_validator_v1:owner_address(Val),
+    LastHeartbeat = blockchain_ledger_validator_v1:last_heartbeat(Val),
+    Stake = blockchain_ledger_validator_v1:stake(Val),
+    Status = blockchain_ledger_validator_v1:status(Val),
+    Version = blockchain_ledger_validator_v1:version(Val),
+    Tenure = maps:get(tenure, Penalties, 0.0),
+    DKG = maps:get(dkg, Penalties, 0.0),
+    Perf = maps:get(performance, Penalties, 0.0),
+    TotalPenalty = Tenure+Perf+DKG,
+    VerboseItems = case Verbose of
```
madninja, line 170: Can you just remove the Verbose option altogether (i..e remove it as an argument to this function too) and just return these items too?

ke6jjj: Absolutely. I didn't like the verbose option anyway.

ke6jjj:
Ok. I believe I have rebased properly. Fingers crossed?

madninja:
Something went wrong with the rebase? It's picking up changes quite unrelated to this PR

ke6jjj:
Closing this PR and reopening from another, re-constructed branch that hasn't been foobar'd.

PR #922, opened 2021-07-12
Port 0 for jsonrpc in test config for run.sh

vihu:
Setting jsonrpc port 0 for being able to do local `run.sh`, otherwise you run into some `eaddrinuse` errors and none but one test miners boot.

### DIFF
--- a/config/test.config
+++ b/config/test.config
@@ -23,7 +23,7 @@
  {miner,
   [
    {mode, validator},
-   {jsonrpc_port, 14464},
+   {jsonrpc_port, 0},
    {use_ebus, false},
    {block_time, 500},
    {election_interval, 10},


Vagabond (approved):

config/test.config
```diff
@@ -23,7 +23,7 @@
  {miner,
   [
    {mode, validator},
-   {jsonrpc_port, 14464},
+   {jsonrpc_port, 0},
```
abhay, line 26: Similar stuff going on here. https://github.com/helium/miner/pull/971/files#diff-50477c121e03cf99d85d5dbdafb9dde09ef25697b3dd0ee4245ccda787e4dbbeR26 fyi, @evanmcc

PR #923, opened 2021-07-12
Start next round early if node does not have its proposal in

Vagabond:
If a node does not get its proposal into a round have it start the next
round early to get a jump on things. This should help with cases where
the node is behind and, because it has to validate all the blocks it
didn't sign, can't catch up for the whole epoch.

### DIFF
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -415,21 +415,27 @@ handle_info({blockchain_event, {add_block, Hash, Sync, Ledger}},
                                                    blockchain_block:transactions(Block),
                                                    Sync}),
                                 {ok, ConsensusAddrs} = blockchain_ledger_v1:consensus_members(Ledger),
+                                F = (length(ConsensusAddrs) - 1) div 3,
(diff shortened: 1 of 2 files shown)


src/miner.erl
```diff
@@ -413,8 +414,23 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
                                   ConsensusGroup, {next_round, NextRound,
                                                    blockchain_block:transactions(Block),
                                                    Sync}),
-                                set_next_block_timer(State#state{current_height = Height});
-
+                                {ok, ConsensusAddrs} = blockchain_ledger_v1:consensus_members(Ledger),
+                                BBAs = blockchain_utils:bitvector_to_map(length(ConsensusAddrs), blockchain_block_v1:bba_completion(Block)),
+                                BBAAddrs = maps:fold(fun(K, true, Acc) ->
```
evanmcc, line 419: I think that we should check the size of `BBAAddrs` and if it's less than 2f+1 we should not actually go early.  sometimes there are still 0 bba votes, we need to track those down, but when they happen they would cause the whole group to go immediately, causing a short block.

evanmcc (approved):

vihu (approved):

Issue #873, opened 2021-06-29
JSONRPC - hbbft_perf


deasydoesit:
### Issue
- The `hbbft_perf` command on the JSONRPC endpoint returns `Method not found.`

### Expected behavior
- The `hbbft_perf` command on the JSONRPC should return consensus group penalty information in JSON format

### Further info
- Command: 
> ` curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":1,"method":"hbbft_perf"}' http://localhost:4467`
- Return value:
> `{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found."},"id":1}`

evanmcc:
This is not implemented and will need to be implemented.

ke6jjj:
I'll do it.

ke6jjj:
This got merged. Can be closed now.

vihu:
https://github.com/helium/miner/pull/930

PR #909, opened 2021-07-08
Implement hbbft_perf for JSONRPC.

ke6jjj:
Implement the `hbbft_perf` for JSON/RPC use, as requested in
helium/miner #873.

A large portion of this code originally lived entirely within
`miner_cli_hbbft.erl`, so rather than duplicate it, the essential
information gathering now happens in `miner_util.erl` (per Evan's
suggestion) and is then parsed and formatted in different ways
by the CLI (`miner_cli_hbbft.erl`) and the JSON/RPC handler
(`miner_jsronrpc_hbbft.erl`).

PR #930, opened 2021-07-15
Implement hbbft_perf for JSONRPC.

ke6jjj:
Implement the `hbbft_perf` for JSON/RPC use, as requested in
helium/miner #873.

A large portion of this code originally lived entirely within
`miner_cli_hbbft.erl`, so rather than duplicate it, the essential
information gathering now happens in `miner_util.erl` (per Evan's
suggestion) and is then parsed and formatted in different ways
by the CLI (`miner_cli_hbbft.erl`) and the JSON/RPC handler
(`miner_jsronrpc_hbbft.erl`).

This is a cherry-pick commit to rescue a branch that has been
screwed up by a bad rebase/merge.

### DIFF
--- a/src/miner_util.erl
+++ b/src/miner_util.erl
@@ -14,12 +14,15 @@
          mark/2,
          metadata_fun/0,
          random_peer_predicate/1,
-         has_valid_local_capability/2
+         has_valid_local_capability/2,
+         hbbft_perf/0
         ]).
 
 %% get the firmware release data from a hotspot
 -define(LSB_FILE, "/etc/lsb_release").
 -define(RELEASE_CMD, "cat " ++ ?LSB_FILE ++ " | grep RELEASE | cut -d'=' -f2").
+
+-include_lib("blockchain/include/blockchain_vars.hrl").
 
 %%-----------------------------------------------------------------------------
 %% @doc Count the number of occurrences of each element in the list.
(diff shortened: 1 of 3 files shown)


ke6jjj:
This is a re-do of PR #909, which references a branch that appears to be hopelessly screwed up.

src/miner_util.erl
```diff
@@ -136,3 +139,59 @@ has_valid_local_capability(Capability, Ledger) ->
...
+                  SeenVotes = blockchain_block_v1:seen_votes(Blk),
+                  Seen = lists:foldl(
+                           fun({_Idx, Votes0}, Acc) ->
+                                   Votes = blockchain_utils:bitvector_to_map(
+                                             length(ConsensusAddrs), Votes0),
+                                   merge_map(ConsensusAddrs, Votes, H, Acc)
+                           end,SeenAcc, SeenVotes),
+                  {merge_map(ConsensusAddrs, BBAs, H, BBAAcc), Seen, Count + length(SeenVotes)}
+          end, {InitMap, InitMap, 0}, Blocks),
+    {ConsensusAddrs, BBATotals, SeenTotals, TotalCount, GroupWithPenalties, Start0, Start, End}.
```
madninja, line 190: I would personally return this as a map instead of the endlessly long tuple, but this will work for now

madninja (approved):

PR #934, opened 2021-07-16
add block age dbus endpoint

BenoitDuffez:
```
root@klk-wifc-030270:~ # dbus-send --system --print-reply --reply-timeout=120000 --dest=com.helium.Miner / com.helium.Miner.BlockAge
method return time=1626454711.232803 sender=:1.201 -> destination=:1.203 serial=4 reply_serial=2
   int32 174
```

### DIFF
--- a/src/miner_ebus.erl
+++ b/src/miner_ebus.erl
@@ -20,6 +20,7 @@
 -define(MINER_MEMBER_ADD_GW, "AddGateway").
 -define(MINER_MEMBER_ASSERT_LOC, "AssertLocation").
 -define(MINER_MEMBER_P2P_STATUS, "P2PStatus").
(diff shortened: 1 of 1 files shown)


abhay:
Seems fine. Want to add block height as well? (can be a separate PR)

abhay (approved):

abhay:
Spoke to @Vagabond offline. Landing.

BenoitDuffez:
Height already available through P2P status

abhay:
> Height already available through P2P status

ah weird. but okay!

BenoitDuffez:
Re: PR: https://github.com/helium/miner/pull/935 (a tiny bit faster than P2PStatus if one only wants the height).

PR #939, opened 2021-07-18
Clean up JSON/RPC API for proper use of errors and data types.

ke6jjj:
Clean up the JSON/RPC API results so that conditional states, missing data
and errors are all easier to detect programmatically. General principles:

* When there is an error, make sure to consistenly use the `?jsonrpc_error()`
  macro.

* If a value is semantically boolean, make it so in the result set. Don't
  return a string.

* If there's no particular error in a query, but there's no data for an item
  then try to use `null` instead of a specialized strings such as "undefined".

* If a query only returns data if the validator/miner is in a specific mode,
  try to ensure that there is a consistently available attribute that
  can be checked _first_ in the result set. Don't make clients try to guess
  the state by checking for the existence of other items, or even worse,
  by checking the returned datatype.

  Example: `dkg_queue`. Before this change, `dkg_queue` either returned

       { "error" : "not running" } (return type is dictionary)
  or

        [ {"name":...,"address":...}, ... ]` (return type is array)

  Now returns either

      { "running": false }
  or

      { "running": true, "consensus_members" : [ ... ] }

### DIFF
--- a/src/jsonrpc/miner_jsonrpc_dkg.erl
+++ b/src/jsonrpc/miner_jsonrpc_dkg.erl
@@ -13,13 +13,13 @@
 %%
 
 handle_rpc(<<"dkg_status">>, []) ->
-    Status = try
-                 miner_consensus_mgr:dkg_status() of
-                 Stat -> Stat
-             catch _:_ ->
-                       <<"not_running">>
-             end,
-    #{ status => Status };
+    try
+        miner_consensus_mgr:dkg_status() of
+            not_running -> #{ running => false };
+            Status -> #{ running => true, status => Status }
+    catch _:_ ->
+        #{ running => false }
+    end;
 handle_rpc(<<"dkg_queue">>, []) ->
     #{ inbound := Inbound,
        outbound := Outbound } = miner:relcast_queue(dkg_queue),
@@ -62,14 +62,20 @@ handle_rpc(<<"dkg_running">>, []) ->
     Delay = max(0, (Diff div RestartInterval) * RestartInterval),
     case Curr >= Height of
         false ->
-            #{ error => <<"not running">> };
+            #{ running => false };
         _ ->
             {ok, Block} = blockchain:get_block(Height+Delay, Chain),
             Hash = blockchain_block:hash_block(Block),
             ConsensusAddrs = blockchain_election:new_group(Ledger, Hash, N, Delay),
-            [ #{ name => ?BIN_TO_ANIMAL(A),
-                 address => ?BIN_TO_B58(A) }
-              || A <- ConsensusAddrs ]
+            #{
+                running => true,
+                consensus_members => [
+                    #{
+                        name => ?BIN_TO_ANIMAL(A),
+                        address => ?BIN_TO_B58(A)
+                   } || A <- ConsensusAddrs
+                ]
+            }
     end;
 handle_rpc(<<"dkg_next">>, []) ->
     Chain = blockchain_worker:blockchain(),
--- a/src/jsonrpc/miner_jsonrpc_hbbft.erl
+++ b/src/jsonrpc/miner_jsonrpc_hbbft.erl
@@ -58,7 +58,7 @@ handle_rpc(<<"hbbft_perf">>, []) ->
         miner_util:hbbft_perf(),
     [
      #{
-	name => ?TO_VALUE(?TO_ANIMAL_NAME(A)),
+        name => ?TO_VALUE(?TO_ANIMAL_NAME(A)),
         address => ?TO_VALUE(?TO_B58(A)),
         bba_completions => [element(2, maps:get(A, BBATotals)), End+1 - Start],
         seen_votes => [element(2, maps:get(A, SeenTotals)), TotalCount],
--- a/src/jsonrpc/miner_jsonrpc_info.erl
+++ b/src/jsonrpc/miner_jsonrpc_info.erl
@@ -46,7 +46,7 @@ handle_rpc(<<"info_p2p_status">>, []) ->
 handle_rpc(<<"info_region">>, []) ->
     R =
         case miner_lora:region() of
-            {ok, undefined} -> <<"undefined">>;
+            {ok, undefined} -> null;
             {ok, Region} -> atom_to_binary(Region, utf8)
         end,
     #{region => R};
--- a/src/jsonrpc/miner_jsonrpc_ledger.erl
+++ b/src/jsonrpc/miner_jsonrpc_ledger.erl
@@ -24,7 +24,7 @@ handle_rpc(<<"ledger_balance">>, #{<<"address">> := Address}) ->
     try
         BinAddr = ?B58_TO_BIN(Address),
         case blockchain_ledger_v1:find_entry(BinAddr, get_ledger()) of
-            {error, not_found} -> #{Address => <<"not_found">>};
+            {error, not_found} -> ?jsonrpc_error({not_found, Address});
             {ok, Entry} -> format_ledger_balance(BinAddr, Entry)
         end
     catch
@@ -91,7 +91,7 @@ handle_rpc(<<"ledger_validators">>, #{ <<"address">> := Address }) ->
     try
         BinAddr = ?B58_TO_BIN(Address),
         case blockchain_ledger_v1:get_validator(BinAddr, Ledger) of
-            {error, not_found} -> #{ error => not_found };
+            {error, not_found} -> ?jsonrpc_error({not_found, Address});
             {ok, Val} -> format_ledger_validator(BinAddr, Val, Ledger, Height)
         end
     catch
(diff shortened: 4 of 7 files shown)


src/jsonrpc/miner_jsonrpc_hbbft.erl
```diff
@@ -58,7 +58,7 @@ handle_rpc(<<"hbbft_perf">>, []) ->
         miner_util:hbbft_perf(),
     [
      #{
-	name => ?TO_VALUE(?TO_ANIMAL_NAME(A)),
+        name => ?TO_VALUE(?TO_ANIMAL_NAME(A)),
```
ke6jjj, line 61: This is a whitespace fix.

src/jsonrpc/miner_jsonrpc_dkg.erl
```diff
@@ -13,13 +13,13 @@
...
-                       <<"not_running">>
-             end,
-    #{ status => Status };
+    try
+        miner_consensus_mgr:dkg_status() of
+            not_running -> #{ running => false };
+            Status -> #{ running => true, status => Status }
+    catch _:_ ->
+        #{ running => false }
+    end;
```
ke6jjj, line 22: This is a return-type normalization. There will always be a `running` attribute to check.

src/jsonrpc/miner_jsonrpc_dkg.erl
```diff
@@ -62,14 +62,20 @@ handle_rpc(<<"dkg_running">>, []) ->
...
-              || A <- ConsensusAddrs ]
+            #{
+                running => true,
+                consensus_members => [
+                    #{
+                        name => ?BIN_TO_ANIMAL(A),
+                        address => ?BIN_TO_B58(A)
+                   } || A <- ConsensusAddrs
+                ]
+            }
```
ke6jjj, line 78: Another return-type normalization. `running` will always exist. `consensus_members` will show up if `running` is true.

src/jsonrpc/miner_jsonrpc_info.erl
```diff
@@ -46,7 +46,7 @@ handle_rpc(<<"info_p2p_status">>, []) ->
 handle_rpc(<<"info_region">>, []) ->
     R =
         case miner_lora:region() of
-            {ok, undefined} -> <<"undefined">>;
+            {ok, undefined} -> null;
```
ke6jjj, line 49: JSON semantics: use `null` where possible.

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -24,7 +24,7 @@ handle_rpc(<<"ledger_balance">>, #{<<"address">> := Address}) ->
     try
         BinAddr = ?B58_TO_BIN(Address),
         case blockchain_ledger_v1:find_entry(BinAddr, get_ledger()) of
-            {error, not_found} -> #{Address => <<"not_found">>};
+            {error, not_found} -> ?jsonrpc_error({not_found, Address});
```
ke6jjj, line 27: JSON/RPC semantics: use error for error conditions.

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -91,7 +91,7 @@ handle_rpc(<<"ledger_validators">>, #{ <<"address">> := Address }) ->
     try
         BinAddr = ?B58_TO_BIN(Address),
         case blockchain_ledger_v1:get_validator(BinAddr, Ledger) of
-            {error, not_found} -> #{ error => not_found };
+            {error, not_found} -> ?jsonrpc_error({not_found, Address});
```
ke6jjj, line 94: JSON/RPC semantics: use error for error condition.

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -113,7 +113,7 @@ handle_rpc(<<"ledger_variables">>, #{ <<"name">> := Name }) ->
             {ok, Var} ->
                 #{ Name => ?TO_VALUE(Var) };
             {error, not_found} ->
-                #{ error => not_found}
+                ?jsonrpc_error({not_found, Name})
```
ke6jjj, line 116: JSON/RPC semantics: use error for error condition.

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -163,7 +163,7 @@ format_ledger_gateway_entry(Addr, GW, Height, Verbose) ->
             }
     end.
 
-last_challenge(_Height, undefined) -> <<"undefined">>;
+last_challenge(_Height, undefined) -> null;
```
ke6jjj, line 166: JSON/RPC semantics: use `null` where appropriate.

src/jsonrpc/miner_jsonrpc_sc.erl
```diff
@@ -9,15 +9,15 @@
 handle_rpc(<<"sc_active">>, []) ->
     case (catch blockchain_state_channels_server:active_sc_ids()) of
         {'EXIT', _} -> ?jsonrpc_error(timeout);
-        undefined -> #{<<"active">> => <<"none">>};
+        undefined -> #{<<"active">> => null};
```
ke6jjj, line 12: Use `null` where appropriate.

src/jsonrpc/miner_jsonrpc_sc.erl
```diff
@@ -9,15 +9,15 @@
...
         BinIds -> #{<<"active">> => [ ?TO_VALUE(base64:encode(I)) || I <- BinIds ]}
     end;
 handle_rpc(<<"sc_active">>, Params) ->
     ?jsonrpc_error({invalid_params, Params});
 handle_rpc(<<"sc_list">>, []) ->
     case (catch blockchain_state_channels_server:state_channels()) of
         {'EXIT', _} -> ?jsonrpc_error(timeout);
-        undefined -> #{<<"channels">> => <<"none">>};
+        undefined -> #{<<"channels">> => null};
```
ke6jjj, line 20: Use `null` where appropriate.

src/jsonrpc/miner_jsonrpc_snapshot.erl
```diff
@@ -14,7 +14,7 @@
 handle_rpc(<<"snapshot_list">>, []) ->
     Chain = blockchain_worker:blockchain(),
     case blockchain:find_last_snapshots(Chain, 5) of
-        undefined -> #{ error => <<"no snapshots found">> };
+        undefined -> ?jsonrpc_error(no_snapshots_found);
```
ke6jjj, line 17: Use error for error conditions.

src/jsonrpc/miner_jsonrpc_txn.erl
```diff
@@ -12,7 +12,7 @@
 
 handle_rpc(<<"txn_queue">>, []) ->
     case (catch blockchain_txn_mgr:txn_list()) of
-        {'EXIT', _} -> #{ error => <<"timeout">> };
+        {'EXIT', _} -> ?jsonrpc_error(timeout);
```
ke6jjj, line 15: Use error for error condition.

src/jsonrpc/miner_jsonrpc_txn.erl
```diff
@@ -42,7 +42,7 @@ handle_rpc(<<"txn_add_gateway">>, #{ <<"owner">> := OwnerB58 } = Params) ->
             lager:error("Couldn't do add gateway via JSONRPC because: ~p ~p: ~p",
                         [T, E, St]),
             Error = io_lib:format("~p", [E]),
-            #{ <<"error">> => Error }
+            ?jsonrpc_error({error, Error})
```
ke6jjj, line 45: Use error for error condition.

src/jsonrpc/miner_jsonrpc_txn.erl
```diff
@@ -64,7 +64,7 @@ handle_rpc(<<"txn_assert_location">>, #{ <<"owner">> := OwnerB58 } = Params) ->
             lager:error("Couldn't complete assert location JSONRPC because ~p ~p: ~p",
                         [T, E, St]),
             Error = io_lib:format("~p", [E]),
-            #{ <<"error">> => Error }
+            ?jsonrpc_error({error, Error})
```
ke6jjj, line 67: Use error for error condition.

OrmEmbaar:
Thanks for this PR. I have written a [Golang client library](https://github.com/Factoshi/go-helium) for the RPC API and fully support your suggested breaking changes.

src/jsonrpc/miner_jsonrpc_sc.erl
madninja: Would it be more appropriate to return an empty list here? (and anywhere else where a list is returned)?

src/jsonrpc/miner_jsonrpc_sc.erl
madninja: Would it be more appropriate to return an empty list here? (and anywhere else where a list is returned)?

ke6jjj: Yes!

madninja (approved):

src/jsonrpc/miner_jsonrpc_sc.erl
ke6jjj: Fixed in 0c0e68b

PR #940, opened 2021-07-19
Add TPM support

isergieienkov:


Vagabond:
Very interesting, thanks! Do you know how to find a list of TPM modules that TPM2 works with?

Vagabond:
Would it be better to avoid adding a whole extra profile/docker image and simply make this configurable via sys.config or ENV vars?

madninja:
Yeah ideally this would be a configuration entry in sys.config to make the "crypto worker" module configurable

Vagabond:
> Yeah ideally this would be a configuration entry in sys.config to make the "crypto worker" module configurable

This is already done by setting the blockchain key appenv var:

```
{key, {tpm, [{key_path, "HS/SRK/MinerKey"}]}},
```

This boots the TPM crypto worker instead of the ECC worker.

madninja:
> > Yeah ideally this would be a configuration entry in sys.config to make the "crypto worker" module configurable
> 
> This is already done by setting the blockchain key appenv var:
> 
> ```
> {key, {tpm, [{key_path, "HS/SRK/MinerKey"}]}},
> ```
> 
> This boots the TPM crypto worker instead of the ECC worker.

ah! indeed it is.. 👍🏻

isergieienkov:
> > Yeah ideally this would be a configuration entry in sys.config to make the "crypto worker" module configurable
> 
> This is already done by setting the blockchain key appenv var:
> 
> ```
> {key, {tpm, [{key_path, "HS/SRK/MinerKey"}]}},
> ```
> 
> This boots the TPM crypto worker instead of the ECC worker.

If you know how to do it without separate docker file please share.

mrallen1:
Hi - planning to work on this a bit - what I want to deliver is a cleaner integration path for y'alls build requirements in the Docker layer and elsewhere.

isergieienkov:
@jadeallenx is anything preventing the merge left?

Dockerfile
```diff
@@ -38,7 +38,7 @@ FROM ${RUNNER_IMAGE} as runner
 
 ARG EXTRA_RUNNER_APK_PACKAGES
 
-RUN apk add --no-cache --update ncurses dbus gmp libsodium gcc \
+RUN apk add --no-cache --update ncurses dbus gmp libsodium gcc tpm2-tss-esys tpm2-tss-fapi tpm2-tss-mu tpm2-tss-rc tpm2-tss-tcti-device\
```
jadeallenx, line 41: All of these extra packages are supposed to go into the `EXTRA_RUNNER_APK_PACKAGES` build-arg

Dockerfile
```diff
@@ -11,7 +11,7 @@ ARG EXTRA_BUILD_APK_PACKAGES
 RUN apk add --no-cache --update \
     git tar build-base linux-headers autoconf automake libtool pkgconfig \
     dbus-dev bzip2 bison flex gmp-dev cmake lz4 libsodium-dev openssl-dev \
-    sed wget curl \
+    sed wget curl tpm2-tss-dev\
```
jadeallenx, line 14: These extra packages are supposed to go into the `EXTRA_BUILD_APK_PACKAGES` build-arg

.buildkite/scripts/make_image.sh
```diff
@@ -33,12 +33,25 @@ case "$BUILD_TYPE" in
         DOCKER_NAME="${BASE_DOCKER_NAME}-${IMAGE_ARCH}_testnet_${VERSION}"
         LATEST_TAG="latest-val-${IMAGE_ARCH}"
         ;;
+    "val_tpm")
+        echo "Doing a testnet validator with TPM image build for ${IMAGE_ARCH}"
+        DOCKER_BUILD_ARGS="--build-arg BUILDER_IMAGE=$BASE_IMAGE --build-arg RUNNER_IMAGE=$BASE_IMAGE --build-arg REBAR_BUILD_TARGET=docker_testval_tpm $DOCKER_BUILD_ARGS"
+        BASE_DOCKER_NAME="validator_tpm"
```
jadeallenx, line 39: This is where you want to define your extra runner and build packages - should be for each tag or some other mechanism that makes sense.

config/docker-val.config.src
```diff
@@ -18,7 +18,7 @@
    {block_sync_batch_limit, 100},
    {max_inbound_connections, 32},
    {snapshot_memory_limit, 2048},
-   {key, undefined},
+   {key, {tpm, [{key_path, "HS/SRK/MinerKey"}]}},
```
jadeallenx, line 21: Maybe it would make sense here to add an `{M, F, A}` style callback method for key acquisition instead of continuing to add more special cases for all the different ways to load key material. Thinking like `{callback, {Mod, Func, Args}}` would be a nice escape hatch

.buildkite/scripts/make_image.sh
```diff
@@ -33,12 +33,25 @@ case "$BUILD_TYPE" in
...
+        DOCKER_NAME="${BASE_DOCKER_NAME}-${IMAGE_ARCH}_testnet_${VERSION}"
```
jadeallenx, line 40: Also, sorry if this is a stupid question, but do these libraries work/exist on ARM64 as well as AMD64?

isergieienkov: Is it armv7 in Alpine packages? Then yes. However I didn't try to build it on ARM.

Dockerfile
isergieienkov: How I will be able to build TPM mapper library? erltss2. Is there any way to make "deps" in rebar.config conditional? And in app.src.
If it is possible, then it worth to exclude ECC code from RPM build

src/miner.app.src
```diff
@@ -28,6 +28,8 @@
     jsx,
     kvc,
     longfi,
+    public_key,
```
Vagabond, line 31: I suspect public_key is not necessary, unless erltss2 is not properly declaring its own deps.

.buildkite/scripts/make_image.sh
```diff
@@ -33,12 +33,25 @@ case "$BUILD_TYPE" in
...
         DOCKER_BUILD_ARGS="--build-arg BUILDER_IMAGE=${BASE_IMAGE} --build-arg RUNNER_IMAGE=${BASE_IMAGE} --build-arg REBAR_BUILD_TARGET=docker_val ${DOCKER_BUILD_ARGS}"
         BASE_DOCKER_NAME="validator"
         DOCKER_NAME="${BASE_DOCKER_NAME}-${IMAGE_ARCH}_${VERSION}"
         ;;
+    "validator_tpm")
+        echo "Doing a mainnet validator with TPM image build for $IMAGE_ARCH"
+        DOCKER_BUILD_ARGS="--build-arg BUILDER_IMAGE=${BASE_IMAGE} --build-arg RUNNER_IMAGE=${BASE_IMAGE} --build-arg REBAR_BUILD_TARGET=docker_val_tpm ${DOCKER_BUILD_ARGS}"
+        BASE_DOCKER_NAME="validator_tpm"
+        DOCKER_NAME="${BASE_DOCKER_NAME}-${IMAGE_ARCH}_${VERSION}"
+        ;;
```
abhay, line 54: we can probably remove the validator builds though, right? we don't need them any more?

abhay: For the history here, @Vagabond and @jadeallenx. we recommended just trying to build a validator version initially just to get some chain follower running on the freedomfi hardware. but the miner_tpm build seems to be the only one needed now.

jadeallenx: Ah ok. I'm going to pull in this branch to a local fork of miner and beat on it a little bit

PR #956, opened 2021-07-27
bump sibyl and grpcbox deps.  Fix for grpcbox socket restart issue

andymck:
Pulls in a fix for grpcbox issue where grpcbox_socket goes down and cannot restart as another process ( acceptor_pool ) is controlling the listen port.  Ends up breaching supervisor restart strategy, resulting in there being no bound listen port.   Grpcbox is effectively down.

Further context is provided in the commit here:

https://github.com/andymck/grpcbox/commit/fbf689bb9c25fc2943155c891974e1f745ce5ac7


Vagabond:
Have you tried reuseport socket option instead?

andymck:
hmm i had thought the socket was already using that but i confused it with reuseaddr.  I will try it again with the addition of reuseport

madninja (approved):

PR #958, opened 2021-07-28
Add VIP router to node_aliases

michaeldjeffrey:


vihu (approved):

Vagabond:
This should no longer be necessary because we limit the size of the peerbook entry so it can be gossipped/ARPed.

PR #970, opened 2021-08-09
Support seed debian builds

mrallen1:
Problem to solve: We want to build debs for seed nodes to make installing/upgrading seed releases easier across a fleet of seed nodes.

### DIFF
--- a/.buildkite/scripts/packagecloud_upload.sh
+++ b/.buildkite/scripts/packagecloud_upload.sh
@@ -2,13 +2,20 @@
 
 set -euo pipefail
 
-TAG=$( git describe --abbrev=0 --tags | sed -e s/validator// )
+TAG=$( git describe --abbrev=0 --tags | sed -e s/$1// )
 
-PKGNAME="validator_${TAG}_amd64.deb"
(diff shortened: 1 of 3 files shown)


.buildkite/scripts/packagecloud_upload.sh
```diff
@@ -2,13 +2,20 @@
...
-PKGNAME="validator_${TAG}_amd64.deb"
+PKGNAME="$1_${TAG}_amd64.deb"
 
 buildkite-agent artifact download ${PKGNAME} .
 
+if [[ "$1" == "validator" ]]; then
+    # it's _validatorS_ in packagecloud
+    REPO="$1s"
+else
+    REPO="$1"
```
madninja, line 15: why not do the same with seedS?

madninja (commented):
Looks good? Just one nit and I guess

* A question if there's a way to combine build kite seed/validator build steps since they're basically the same

.buildkite/scripts/packagecloud_upload.sh
mrallen1: yeah i guess i should

PR #972, opened 2021-08-09
Switch to CNAME for seed node DNS

Vagabond:


config/sys.config
```diff
@@ -51,7 +53,7 @@
    {onboarding_dir, "/mnt/uboot"},
    {num_consensus_members, 16},
    {seed_nodes, "/ip4/35.166.211.46/tcp/2154,/ip4/44.236.95.167/tcp/2154,/ip4/3.34.61.168/tcp/2154"},
-   {seed_node_dns, "seed.helium.foundation"},
+   {seed_node_dns, ""},
```
vihu, line 56: Can we remove this?

vihu (approved):

abhay:
Do we need to update this here? https://github.com/helium/miner/blob/master/config/seed.config#L13-L14

mrallen1:
> Do we need to update this here? https://github.com/helium/miner/blob/master/config/seed.config#L13-L14

That config only gets incorporated during `make seed`

PR #981, opened 2021-08-14
Update validator libp2p to use DNS for seeds

abhay:


abhay:
Not needed. these get pulled in from the base `sys.config`

PR #983, opened 2021-08-16
Update config files to use listen_addresses

PaulVMo:
This change updates the sys.config files to use the 'listen_addresses' parameter in place of 'port' and 'ports'. This allows the listening IP addresses (not just the listen ports) to be specified. This is a community requested feature to provide better control when running the miner on multi-homed servers.

This PR is dependent on (blockchain-core PR 947) [https://github.com/helium/blockchain-core/pull/947] which includes the changes to use this parameter.


madninja (commented):
Are you missing some code changes to go with this? How is listen_addresses actually picked up and used?

PR #991, opened 2021-08-19
Switch Australia to AS923_1

vihu:
Australia appears to operate on AS923_1 instead of AU915, this updates the CSV we use.

PS: External [hplans](https://github.com/gradoj/hplans/commit/77770ed444d0791f0a8620a7d8eb0c57806ca03a) repository has already been update and so is [lorawan-h3](https://github.com/helium/lorawan-h3/commit/9637525903c7f7b061e0c8ac6fbe7cdc665fdf36).


shawaj:
@vihu where did you get this info from? AU915 is the specific one for Aus and NZ hence the AU in the name?

jmarcelino:
I'm curious where this is coming from too as it may impact existing hardware certifications. These changes shouldn't be done so lightly.

shawaj:
FYI this is the latest region doc... https://lora-alliance.org/resource_hub/rp2-1-0-3-lorawan-regional-parameters/

Which says AU915 for both Aus and NZ

vihu:
Thank you for bringing this up @jmarcelino @shawaj.

We've had some internal discussions about the region plan for Australia and as you both stated, sticking with AU915 for the time being is the right thing to do here. I will close this for now and do the necessary reverts in other related repositories. We will investigate further if AS923_1 is required in the future.

cc: @gradoj

abhay:
Thanks for closing the PR, @vihu. 

Agree with closing this for now. The miner config (today in csv but soon a chain variable on the network after PoCv11) is complex because it's used in both PoC and device traffic. As both @shawaj and @jmarcelino correctly stated, it's not a light change. 

For places that have less regulatory clarity or where device manufacturers on the ground are pushing for new standards, we'll have to react to the changing landscape as a network. We have a method to do this via chain variables. Today these are controlled at a country level but could certainly be done in a smaller region, if someone is willing to do the h3 polygon work.

Also any gateway manufacturer can decide to use an alternate frequency plan (if they choose). They will be risking not being able to PoC with other hotspots in the area but they'll still be able to support the devices / frequency plan they intend to support. This should support frontier deployments if they decide to go another path. 

If there are follow ups on this topic, I think opening a new [issue](https://github.com/helium/miner/issues) is the right call. For now, we stay the course with AU915 in Australia.

shawaj:
Maybe makes sense for decisions like this to be taken into something like the DeWi technical standards or something so that there's more visibility for big changes like this so manufacturers can be more aware?

abhay:
> Maybe makes sense for decisions like this to be taken into something like the DeWi technical standards or something so that there's more visibility for big changes like this so manufacturers can be more aware?

Agree, generally. With pocv11, regions will be chain vars that can be managed through governance

shawaj:
That's great news @abhay 

It doesn't really matter from a hardware/software perspective but as you say it does from a user perspective - i.e. if someone has bought an AU915 hotspot and then it changes to AS923_1 in future and the specific manufacturer won't/can't update this... Could leave them in a bit of a mess and/or with two rival frequency networks within a single country.

Other thing that might be useful is to enforce as part of the hardware audit that the vendors have to get the region plan from the miner/location assert instead of hard coded. I don't really see a reason why you wouldn't?

mdaviesnz:
@jmarcelino Just for some further information on the subject from a NZ local (I'm not sure about Australia on this) but AS923 is preferred and in use by local operators here because there's actually a power limit of 0 dBw eirp (1w) between 915 and 920 MHz. 
Specific regulation here: https://gazette.govt.nz/notice/id/2016-go446
Discussion by a local provider (who actually ended up using IN865) here: https://kotahi.net/radio-frequencies-bands-and-inspections/

I'm not really sure how much this power limit would impact usage (presuming it's actually adhered to) here but I imagine it's not helpful. 
It might be a similar situation in Australia as we tend to follow them on electrical standards but I'm not sure where to look.

I notice that the regional parameters doc lists AS923 and AU915-928 for both NZ and Aus.

Practically if there were ever going to be a time to make changes like this it would be while the network is young and there's not massive amounts of devices/sensors that'd need changing. Once it's confirmed whether the hotspot vendors can adjust the frequency plans on deployed hotspots (seems Bobcat could be the laggard here) it might not be too hard to make a country-wide change if required (and beneficial).

abhay:
Once PoCv11 is landed, it'll be an on chain/governance decision to change regional parameters. This will be driven by the Foundation as they are now official members of the LoRa Alliance.

PR #714, opened 2021-03-30
sokhouw/witness random sampling

sokhouw:
This is proposed implementation for issue [#711](https://github.com/helium/miner/issues/711)

PR #1026, opened 2021-09-02
use a random selection of witnesses instead of the first LIMIT

evanmcc:
I think that using a bloom filter as per #714 is both overcomplicated and biases witness selection in an odd way.  I'd like to try this first; This PR simply accumulates all of the witnesses and takes a random slice of them, which should be fairer.

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -320,22 +320,16 @@ receiving(cast, {witness, Address, Witness}, #data{responses=Responses0,
                     {keep_state, Data};
                 _ ->
                     Witnesses = maps:get(PacketHash, Responses0, []),
-                    PerHopMaxWitnesses = blockchain_utils:poc_per_hop_max_witnesses(Ledger),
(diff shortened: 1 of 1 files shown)


ke6jjj:
@evanmcc I'm happy to see that this is being explored! I think I'm the instigator who suggested using the Bloom filter. I did so because I was very worried about possible Denial-of-Service attacks. If you keep a straight list of previous reports, it's possible for an adversary to submit reams and reams of reports in an attempt to shoot down the whole Challenge.

evanmcc:
@ke6jjj I'm happy to try and mitigate that in other ways.  At the moment, we throw away any repeated receipts. I think that we also check that the witness exists?  So in order to DoS, you'd need to sign a lot of stuff that someone could potentially submit as proof of bad actions in the future.

Perhaps I'm not thinking this through fully, I'll give the validation another look.

Vagabond (approved):
I think this is fine and is not particularly worse from a DoS perspective than anything we had before.

PR #1036, opened 2021-09-07
Handle and test transaction protocol v2

xandkar:
Which expects a node's height to be passed along with the transaction acceptance status.

Pre-merge TODO:
- [x] merge https://github.com/helium/blockchain-core/pull/983
- [ ] after merging above, update `rebar.lock`, so it points to core with v2 support
- [x] uncomment `cleanup_per_testcase` in `miner_ct_utils`

### DIFF
--- a/src/miner_hbbft_sidecar.erl
+++ b/src/miner_hbbft_sidecar.erl
@@ -146,6 +146,7 @@ handle_call({submit, Txn}, From,
                             case blockchain_txn:absorb(Txn, Chain) of
                                 ok ->
                                     spawn(fun() ->
+                                        lager:warning("absorbed ok. spawning relcast handle_command txn: ~p", [blockchain_txn:print(Txn)]),
                                                 catch libp2p_group_relcast:handle_command(Group, {txn, Txn})
                                         end),
                                     {reply, {ok, Height}, State};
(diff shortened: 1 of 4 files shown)


Vagabond:
You need to handle the noreplys chained off the sidecar submission event that are later replied to with gen_server:reply to also contain the chain height.

test/miner_ct_utils.erl
```diff
@@ -964,7 +964,8 @@ end_per_testcase(TestCase, Config) ->
     case ?config(tc_status, Config) of
         ok ->
             %% test passed, we can cleanup
-            cleanup_per_testcase(TestCase, Config);
+            %cleanup_per_testcase(TestCase, Config);
```
Vagabond, line 967: TODO: remove?

xandkar: Yeah. Did not mean to commit... oops

src/miner_hbbft_sidecar.erl
```diff
@@ -91,20 +96,21 @@ handle_call({set_group, Group}, _From, #state{group = OldGroup} = State) ->
         {P1, P2} when is_pid(P1) andalso is_pid(P2)  ->
             ok;
         {undefined, P} when is_pid(P) ->
-            ok = libp2p_swarm:add_stream_handler(blockchain_swarm:tid(), ?TX_PROTOCOL,
+            ok = libp2p_swarm:add_stream_handler(blockchain_swarm:tid(), ?TX_PROTOCOL_V2,
```
Vagabond, line 99: You need to install/uninstall both protocol versions

src/miner_hbbft_sidecar.erl
```diff
@@ -91,20 +96,21 @@ handle_call({set_group, Group}, _From, #state{group = OldGroup} = State) ->
...
                                                  {libp2p_framed_stream, server,
                                                   [blockchain_txn_handler, self(),
```
Vagabond, line 101: You may need to supply the protocol version here so it shows up in the server side's argument list.

PR #1040, opened 2021-09-08
compile in pid file

densone:
## brief
for the folks out there wanting to run validators outside of docker it would be very useful to have a pid file in place. This will allow us to utilize tools like sysrc more effectively and allow us to opensource upstart scripts.

## This has been compiled and tested with the latest validator beta branch 

```bash
root@casey:~/miner/_build/validator/rel/miner # cat miner.pid
95634
root@casey:~/miner/_build/validator/rel/miner # kill `cat miner.pid`
root@casey:~/miner/_build/validator/rel/miner # kill `cat miner.pid`
95634: No such process

root@casey:~/miner/_build/validator/rel/miner # bin/miner daemon
root@casey:~/miner/_build/validator/rel/miner # cat miner.pid
98007
root@casey:~/miner/_build/validator/rel/miner # kill `cat miner.pid`

```

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -81,7 +81,8 @@
          [
           {post_start,
            [
-            {wait_for_process, blockchain_worker}
+            {wait_for_process, blockchain_worker},
+            {pid, "miner.pid"}
            ]}
          ]},
         {extended_start_script, true},


rebar.config
```diff
@@ -81,7 +81,8 @@
          [
           {post_start,
            [
-            {wait_for_process, blockchain_worker}
+            {wait_for_process, blockchain_worker},
+            {pid, "miner.pid"}
```
evanmcc, line 85: while it's nice to have a pidfile, shouldn't it be somewhere else, like `/tmp/`?

madninja:
If you use systemd (brr) it creates the pid file for you already. And perhaps that's where it could/should live, i.e. in the place that kicks it all of?

Vagabond:
Maybe this is a good use for rebar.config.script?

rebar.config
densone: depends on the os, but I think tmp is a great place.

madninja:
My point was that the system runner (i.e. systemd) determines where the pid goes

ke6jjj:
@densone maybe taking a tack where this could integrate with `daemon(8)` might work better with this .. ahem, surly crew? 😂

ke6jjj:
> My point was that the system runner (i.e. systemd) determines where the pid goes @densone

Let's say that one does this integration via `systemd` and decides to use `bin/miner daemon` as the launch point. Would `systemd` pick up the correct PID of the miner itself? Or would it pickup the pid of the shell that _invokes_ the miner?

rebar.config
ke6jjj: `/var/run` is canonical on BSD now. So, there are quite a few disagreeable options.

madninja:
> > My point was that the system runner (i.e. systemd) determines where the pid goes @densone
> 
> @madninja: Let's say that one does this integration via `systemd` and decides to use `bin/miner daemon` as the launch point. Would `systemd` pick up the correct PID of the miner itself? Or would it pickup the PID of the shell that _invokes_ the miner?

Using daemon would indeed change the game a little, look forward to the PR to make that happen since the systemd would still need to be told where to create or look for the pid file, no?

systemd makes anyone surly :-)

evanmcc:
I don't think that we want to do this at this level

PR #1042, opened 2021-09-08
fix #1041

### LINKED ISSUE #1041: lost pocs
Hi all,

I was digging through my logs recently and noticed I had a number of sequences that looked like:

```
[info] <0.1576.0>@miner_poc_statem:send_onion:977 onion sent
[info] <0.32135.2>@miner_onion_handler:handle_info:78 client sending data: <<...>>
[warning] <0.1576.0>@miner_poc_statem:handle_event:435 ignoring event [info] {#Ref<0.432738089.1828978691.190885>,ok}
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.32135.2>@miner_onion_handler:handle_data:60 client got data: <<...>>
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.1576.0>@miner_poc_statem:receiving:289 got block <<...>> decreasing timeout
[info] <0.1576.0>@miner_poc_statem:receiving:337 got receipt {blockchain_poc_receipt_v1_pb,<<...>>,0.0,0.0,0,[],<<>>,0}
[warning] <0.1576.0>@miner_poc_statem:handle_event:435 ignoring event [cast] {witness,<<...>>}}
[warning] <0.1576.0>@miner_poc_statem:handle_event:435 ignoring event [cast] {witness,<<...>>}}
[warning] <0.1576.0>@miner_poc_statem:handle_event:435 ignoring event [cast] {witness,<<...>>}}
```

From digging into the code, it seemed this likely happened due to a state change at:

https://github.com/helium/miner/blob/d5a42e7eea1ddd81cfcd4fd21a7bbc32b94ade33/src/poc/miner_poc_statem.erl#L370

When this state change happens, it seems the subsequent witnesses are ignored and the responses are never submitted. At first I was curious if the bloom filter used inside of `check_addr_hash` was producing false positives or something, but then I noticed this line:

https://github.com/helium/miner/blob/d5a42e7eea1ddd81cfcd4fd21a7bbc32b94ade33/src/poc/miner_poc_statem.erl#L790

Not an Erlang dev, but this seems to be calculating the hash using only the first 3 octets of the peer's IP address (the challengee in this case) rather than the full IP address which seemed suspect. Further digging showed that this change came as part of a commit with a commit message of `Bump snapshot to 943201`:

https://github.com/helium/miner/commit/5b4229b63d21f97e2c196d551e634ed21bd00613

Previous to this, it was using the full IP address. Given the commit message made no mention of this change, I wondered if this change was intentional, if I was misunderstanding the code, or if this was just some local change that accidentally sneaked itself in. Pinging @Vagabond since you were the author on the commit :)

ci-work:
quick fix for #1041 - removes 3 octet limit on ip address hash construction

### DIFF
--- a/src/poc/miner_poc_statem.erl
+++ b/src/poc/miner_poc_statem.erl
@@ -787,9 +787,8 @@ check_addr_hash(PeerAddr, #data{addr_hash_filter=#addr_hash_filter{byte_size=Siz
     case multiaddr:protocols(PeerAddr) of
         [{"ip4",Address},{_,_}] ->
             {ok, Addr} = inet:parse_ipv4_address(Address),
-            Val = binary:part(enacl:pwhash(list_to_binary(lists:sublist(tuple_to_list(Addr), 3)), binary:part(Hash, {0, enacl:pwhash_SALTBYTES()})), {0, Size}),
-            Val2 = binary:part(enacl:pwhash(list_to_binary(tuple_to_list(Addr)), binary:part(Hash, {0, enacl:pwhash_SALTBYTES()})), {0, Size}),
-            case bloom:check_and_set(Bloom, Val) orelse bloom:check_and_set(Bloom, Val2) of
+            Val = binary:part(enacl:pwhash(list_to_binary(tuple_to_list(Addr)), binary:part(Hash, {0, enacl:pwhash_SALTBYTES()})), {0, Size}),
+            case bloom:check_and_set(Bloom, Val) of
                 true ->
                     true;
                 false ->


Vagabond:
Ah, whoops, this was a git commit -a snafu.

Vagabond (approved):

leyuansugar:
Within a month of this bug fix the lost poc bug was resolved, but since November, have started experiencing the same situation again and would like to check if there is a new problem?

PR #1050, opened 2021-09-10
Inadvertantly duplicated DOCKER_NAME

jadeallenx:
Problem to solve: When refactoring the script that drives Docker builds, the `$DOCKER_NAME` was inadvertently duplicated twice which led to duplicated tags on the `miner` build target.


JayKickliter (approved):
Thanks. Go ahead an merge when ready

PR #1059, opened 2021-09-14
Handle discovery mode packet forwarder tx timeout more gracefully.

### LINKED ISSUE #1058: Dicovery Mode Worker Crashing
After initiating discovery mode, the worker will crash. 
```
2021-09-14 18:35:45.994 520 [info] <0.24174.7>@miner_discovery_handler:handle_data:59 got <<NUMNBERS>> {state}
2021-09-14 18:35:45.996 520 [info] <0.24176.7>@miner_discovery_worker:init:45 starting discovery mode with 10 packets every 5000 ms
2021-09-14 18:35:50.997 520 [info] <0.2149.1>@miner_lora:send_packet:758 PULL_RESP to {127,0,0,1}:1680
2021-09-14 18:35:53.491 520 [info] <0.2149.1>@miner_lora:handle_udp_packet:390 PULL_DATA from 00000000000000000000 on 1680
2021-09-14 18:36:00.998 520 [error] <0.24176.7>@miner_discovery_worker:handle_info:78 gen_server miner_discovery_worker terminated with reason: no match of right hand value {error,timeout} in miner_discovery_worker:handle_info/2 line 78
2021-09-14 18:36:00.999 520 [error] <0.24176.7>@miner_discovery_worker:handle_info:78 CRASH REPORT Process miner_discovery_worker with 0 neighbours crashed with reason: no match of right hand value {error,timeout} in miner_discovery_worker:handle_info/2 line 78
```
This is after getting the discovery mode, the miner will run it, and run into an timeout error. This timeout error come from: [https://github.com/helium/miner/blob/b756afb438c6ce7f1ca5fd320a5704d9b395fb5f/src/miner_lora.erl#L368](https://github.com/helium/miner/blob/b756afb438c6ce7f1ca5fd320a5704d9b395fb5f/src/miner_lora.erl#L368)
And goes to: [https://github.com/helium/miner/blob/b756afb438c6ce7f1ca5fd320a5704d9b395fb5f/src/miner_discovery_worker.erl#L78](https://github.com/helium/miner/blob/b756afb438c6ce7f1ca5fd320a5704d9b395fb5f/src/miner_discovery_worker.erl#L78)

This is because the there is no case to handle the {error, timeout} from a lora gateway, which happens from time to time, do the lack of ack's by the gateway.

ke6jjj:
It's possible, though usually rare, that the request to transmit a discovery mode beacon packet might time out. When this happens, make sure not to crash the entire discovery mode session but simply move on to the next packet.

Should fix #1058, discovered by @maco2035

### DIFF
--- a/src/miner_discovery_worker.erl
+++ b/src/miner_discovery_worker.erl
@@ -75,13 +75,22 @@ handle_info(
         lists:nth(rand:uniform(length(FreqList)), FreqList)
     end,
     Spreading = spreading(Region, byte_size(Packet)),
-    ok = miner_lora:send_poc(
+    case miner_lora:send_poc(
         Packet,
         immediate,
         ChannelSelectorFun,
         Spreading,
         TxPower
-    ),
+    ) of
+        ok ->
+            ok;
+        {warning, _} ->
+            %% Most warnings should be ok.
+            ok;
+        {error, timeout} ->
+            %% This transmission timed out, but others may work. Keep going.
+            lager:warning("transmission timed out.")
+    end,
     timer:send_after(?DEFAULT_TRANSMIT_DELAY_MS, self(), tick),
     {noreply, State#state{packets = Packets}};
 handle_info(_Info, State) ->


JayKickliter (changes requested):
Can you check if handling `{warning, any()}` is needed here?

ke6jjj:
Sure! I can try to statically analyze if that's possible. There are some other `{error,` cases that can arise, but they all seem unrecoverable and crash worthy.

JayKickliter:
Ok. I trust your judgment. Add another comment when you decide one way or the other and I'll approve

ke6jjj:
It's actually a good thing to visit. I get where you're coming from; `send_poc` is marked as returning `{warning, any()}` in the spec. And sure enough, maybe we should also handle the possibility that the transmit power was reduced:

`{warning, {tx_power_corrected, ...}}`

JayKickliter:
I can certainly imagine a `tx_power_corrected` bug creeping in in the future

ke6jjj:
Updated to handle `{warning, _}` as well!

JayKickliter (approved):

Vagabond (approved):

PR #1079, opened 2021-09-23
Jma/tpm

jadeallenx:
Problem to solve: We want to make TPM integration generally optional to the main miner build pathways but also be able to support TPM builds.

See also #940


Vagabond:
can you see how this looks for you @isergieienkov ?

isergieienkov:
> can you see how this looks for you @isergieienkov ?

Yes, TPM miner is working just fine.

PR #1081, opened 2021-09-23
Add poc-v11 var proposal and complete var file

vihu:
Draft PoC-v11 chain variable proposal

### DIFF
--- a/audit/var-poc-v11.md
+++ b/audit/var-poc-v11.md
@@ -25,7 +25,7 @@ region_ru864
 region_us915
 ```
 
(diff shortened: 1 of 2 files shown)


audit/var-poc-v11.md
```diff
@@ -0,0 +1,51 @@
...
+region_cn470
+region_eu433
+region_eu868
+region_in865
+region_kr920
+region_ru864
+region_us915
+```
+
+* For each of the above `region_*` variables there will be a corresponding `region_*_params` chain variable which determines the specific LoraWAN region parameters a device is allowed to operate legally. These are binary blobs with encoding driven using protobuf available [here](https://github.com/helium/proto/blob/master/src/blockchain_region_param_v1.proto). A deserialized version of the below variables is available [here](https://gist.github.com/vihu/ac2d6761318672376db7a12fa915519c).
```
abhay, line 28: can you move the gist to a checked in file as well?

PR #1101, opened 2021-09-28
Testnet 

tzacha:
Was missing support for 'region_as923_4'  in region types and models, 
so returned ' is not a supported regulatory region'
this affects all miners in Israel and needs resolution ASAP


tzacha:
@vihu can you please review my update?
only added region_as923_4 in 2 lines
it is very urgent for me to release it. 

thanks in advance

vihu:
Thanks, we already got it merged https://github.com/helium/miner/pull/1100

PR #1119, opened 2021-10-07
Fix miner_jsonrpc_sc to include new SC functions

macpie:


src/jsonrpc/miner_jsonrpc_sc.erl
```diff
@@ -7,18 +7,18 @@
 %% jsonrpc_handler
 %%
 handle_rpc(<<"sc_active">>, []) ->
-    case (catch blockchain_state_channels_server:active_sc_ids()) of
-        {'EXIT', _} -> ?jsonrpc_error(timeout);
-        undefined -> #{<<"active">> => []};
-        BinIds -> #{<<"active">> => [ ?TO_VALUE(base64:encode(I)) || I <- BinIds ]}
+    case (catch blockchain_state_channels_server:get_actives()) of
+        {'EXIT', _} ->
```
jeffgrunewald, line 11: (making sure I understand what's going on here) are we catching this call b/c this is part of the RPC public API and the calling process is trapping exits (hence the first `case` clause) but we're not catching the same call below in `format_sc_map/1` its an internal function?

src/jsonrpc/miner_jsonrpc_sc.erl
```diff
@@ -28,12 +28,20 @@ handle_rpc(_, _) ->
...
-              end, [], SCs).
+    ActiveSCIDs = maps:keys(blockchain_state_channels_server:get_actives()),
+    maps:fold(
+        fun(SCID, SC, Acc) ->
+            #{expire_at_block := ExpireAt} = Json = blockchain_state_channel_v1:to_json(SC, []),
+            [
+                Json#{
+                    <<"is_active">> => lists:member(SCID, ActiveSCIDs),
+                    <<"expired">> => Height >= ExpireAt
+                }
```
jeffgrunewald, line 41: is this the Erlang equivalent Map merging syntax to Elixir's:
```elixir
%{
  json |
  "is_active" => Lists.member(scid, active_scids),
  "expired" => height >= exire_at
}
```

jeffgrunewald: nvm, I've confirmed this; it's actually _more_ powerful than Elixir's map update shorthand syntax b/c there it's not supported to use that syntax if the key isn't already present in the map (only equivalent to `Json#{ <<"is_active">> := lists:member(SCID, ActiveSCIDs)}`) which is pretty cool!

Issue #1132, opened 2021-10-16
miner_lora sends invalid immediate packets


timcooijmans:
The miner sends the string `immediate` in the tmst-field of the LoRa UDP protocol for immediate packets while this field is defined as a number. 
`{"txpk":{"codr":"4/5","data":"<redacted>","datr":"SF12BW125","freq":868.5,"imme":true,"ipol":false,"modu":"LORA","powe":14,"rfch":0,"size":52,"tmst":"immediate"}}`

Fix is to set the field to 0 when sending immediate packets.

PR #1136, opened 2021-10-18
Ensure that tmst is an integer for all packets

vihu:
Potential fix for https://github.com/helium/miner/issues/1132.


timcooijmans:
Yes, this works. The tmst field is ignored when `imme=true` so the value doesn't really matter. It was the wrong type that caused our issues, and this fixes that.

PR #1158, opened 2021-11-03
Add a JSON-RPC query for LoRa region params.

ke6jjj:
Add a JSON-RPC endpoint for querying LoRa region parameters to support manufacturer integration with PoC v11's dynamic (on-chain) region parameterization.

### DIFF
--- a/src/jsonrpc/miner_jsonrpc_ledger.erl
+++ b/src/jsonrpc/miner_jsonrpc_ledger.erl
@@ -121,6 +121,28 @@ handle_rpc(<<"ledger_variables">>, #{ <<"name">> := Name }) ->
     end;
 handle_rpc(<<"ledger_variables">>, Params) ->
     ?jsonrpc_error({invalid_params, Params});
+handle_rpc(<<"ledger_lora_region_parameters">>, #{ <<"region">> := Region}) ->
+    Ledger = get_ledger(),
+    try
+        RegionAtom = binary_to_existing_atom(Region, utf8),
+        case blockchain_region_params_v1:for_region(RegionAtom, Ledger) of
+            {ok, RegionParams} ->
+                format_region_parameters(RegionParams);
+            {error, _Reason} ->
+                ?jsonrpc_error({not_found, Region})
+        end
+    catch
+        %% String-to-atom conversion will fail with this error if there is no
+        %% atom present in the runtime for the given region. This is an
+        %% implicit signal that there are no parameters for the region, and
+        %% thus, the region likely doesn't exist.
+        error:badarg ->
+            ?jsonrpc_error({not_found, Region})
+    end;
+handle_rpc(<<"ledger_lora_region_parameters">>, []) ->
+    ?jsonrpc_error({invalid_params, <<"region required">>});
+handle_rpc(<<"ledger_lora_region_parameters">>, Params) ->
+    ?jsonrpc_error({invalid_params, Params});
 handle_rpc(_, _) ->
     ?jsonrpc_error(method_not_found).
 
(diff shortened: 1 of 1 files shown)


madninja (approved):
that looks very nice.. well done erlang @ke6jjj

ke6jjj:
When successful, this query returns a result which contains a JSON array of _Channel Objects_.

# Channel Object

|Attribute|JSON type|Description|
|-|-|-|
|`channel_frequency_hz`|number|Channel center frequency, in hertz.|
|`bandwidth_hz` | number | Channel bandwidth, in hertz.|
|`max_eirp_cbm` | number | Maximum permissible Effective Isotropic Radiated Power (EIRP), in centidecibels milliwatt (cBm). Divide by 10 to convert to decibels milliwatt (dBm).|
|`spreading` | array | List of spreading factors used on the channel and their maximum packet sizes. (See SpreadingFactor object).|

# SpreadingFactor object

|Attribute|JSON type|Description|
|-|-|-|
|`factor` | string | Enumeration (one of) `SF12`, `SF11`, `SF10`, `SF9`, `SF8`, `SF7`; representing a LoRa spreading factor.|
|`max_packet_size` | number | Maximum permissible transmit packet size at this spreading factor, in bytes.|

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -189,3 +211,40 @@ format_ledger_validator(Addr, Val, Ledger, Height) ->
...
+    BandwidthHz = blockchain_region_param_v1:bandwidth(RegionParam),
+    MaxEIRPcBm = blockchain_region_param_v1:max_eirp(RegionParam),
+    Spreading = blockchain_region_param_v1:spreading(RegionParam),
+    SpreadingObject = format_spreading(Spreading),
+    #{
+        <<"channel_frequency_hz">> => ChannelFrequencyHz,
+        <<"bandwidth_hz">> => BandwidthHz,
+        <<"max_eirp_cbm">> => MaxEIRPcBm,
+        <<"spreading">> => SpreadingObject
+    }.
```
jeffgrunewald, line 233: for typespecs on functions that return relatively small maps that don't change frequently (or complex tuples for that matter) do we have style guidelines for when we might want a function like this to define a spec like
```erlang
-spec format_region_parameter(blockchain_region_param_v1:region_param_v1()) -> #{binary() => any()}
```
or would we generally prefer to limit this type of specificity to functions that return records?
In this case, I can see the possibility of the `undefined` atom as potential return values for the fields making the map cumbersome so if there's no sensible defaults for region params it probably doesn't make sense for this map.

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -121,6 +121,28 @@ handle_rpc(<<"ledger_variables">>, #{ <<"name">> := Name }) ->
...
+            {error, _Reason} ->
+                ?jsonrpc_error({not_found, Region})
+        end
+    catch
+        %% String-to-atom conversion will fail with this error if there is no
+        %% atom present in the runtime for the given region. This is an
+        %% implicit signal that there are no parameters for the region, and
+        %% thus, the region likely doesn't exist.
+        error:badarg ->
+            ?jsonrpc_error({not_found, Region})
```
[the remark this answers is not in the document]
ke6jjj: Good question! I know that `not_found` seems to be a JSON-RPC standard. Earlier I had a better error message planned, but JSON-RPC was having none of it:

`{no_such_region, Region}`

This caused an internal error and nothing got reported to the user.

src/jsonrpc/miner_jsonrpc_ledger.erl
ke6jjj: I'm still new to Erlang type specs and would love more guidance here. I didn't know how deep to go, nor how to specify a map that had specific keys and the types for those keys.

jeffgrunewald: It's largely a matter of preference and documentation if we're not consistently running dialyzer, particularly as a regular CI check but the example above is a middle ground between `map()` and a fully explicit type definition. With a map that has static contents you could then guard even further against changes that deviate from that spec with `#{<<"channel_frequency_hz">> => undefined | non_neg_integer(), <<"bandwidth_hz">> => undefined |non_neg_integer(), <<"max_eirp_cbm">> => undefined | non_neg_integer(), <<"spreading">> => undefined | blockchain_region_spreading_v1:region_spreading_v1()}` but the possibility of those `undefined`s makes that level of specificity less valuable in my opinion

src/jsonrpc/miner_jsonrpc_ledger.erl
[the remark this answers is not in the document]
madninja: Leave the not_found error in place.. It's the convention used.

src/jsonrpc/miner_jsonrpc_ledger.erl
madninja: I like @jeffgrunewald 's middle ground.. Over-speccing, generally takes a lot of time for little benefit

ke6jjj: I'm not sure what the middle-ground would look like. How is this?

```erlang
-type lora_spreading_spec() :: #{
        <<"factor">> => binary(),
        <<"max_packet_size">> => non_neg_integer()
    }.
-type lora_channel_spec() :: #{
        <<"channel_frequency_hz">> => non_neg_integer(),
        <<"bandwidth_hz">> => non_neg_integer(),
        <<"max_eirp_cbm">> => integer(),
        <<"spreading">> => [lora_spreading_spec()]
    }.
```

ke6jjj: Oh, I see the middle ground now. Well, as you can see, a fully-spec'd type for these isn't that bad.

abhay: +1 on middle ground.

abhay (approved):

src/jsonrpc/miner_jsonrpc_ledger.erl
ke6jjj: Well, my example above apparently isn't a legal type spec. I don't know how to spec a map with specific keys; I can't figure it out from the general BNF for type specs.

jeffgrunewald: That’s right; I recall you can only use explicit map keys in a spec if they’re atoms. `binary()` is probably the most specific binary-type keys would support

ke6jjj: Ok, how about the latest change? I introduced a type called `json_object()` which is nothing more than `#{ binary() => any() }.`

jeffgrunewald (approved):

src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -10,6 +10,8 @@
 %% jsonrpc_handler
 %%
 
+-type json_object() :: #{ binary() => any() }.
```
madninja, line 13: I would not recommend doing this..there is 1. already a type spec for a json object out of jsx I believe
and 2. you can actually use atoms for keys with the used library.. Other jsonrpc endpoints do this

madninja (changes requested):

src/jsonrpc/miner_jsonrpc_ledger.erl
ke6jjj: This module has already established the convention of using `binary()` for keys. I didn't want to rock the boat. I can try to find the jsx type...

madninja: the module is tiny.. you should have gone with the middle ground if you didn't want to rock the boat

ke6jjj: I think `json_object() :: #{ binary() => any() }.` is the middle ground, no?

jeffgrunewald (approved):

madninja (approved):

PR #1159, opened 2021-11-03
More poc-v11 guard checks

vihu:
Guard lookup for regulatory regions further behind `poc_version > 10` check.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -166,31 +166,45 @@ reg_domain_data_for_addr(Addr, #state{chain=Chain}) ->
         undefined ->
             {error, no_ledger};
         Ledger ->
-            %% check if the poc 11 vars are active yet
-            case blockchain_ledger_v1:find_gateway_location(Addr, Ledger) of
-                {ok, Location} ->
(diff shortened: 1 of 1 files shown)


src/miner_lora.erl
```diff
@@ -166,33 +166,47 @@ reg_domain_data_for_addr(Addr, #state{chain=Chain}) ->
...
-                            end;
-                        {error, regulatory_regions_not_set} ->
-                            case country_code_for_addr(Addr) of
-                                {ok, CC} ->
-                                    %% use country code to get regulatory domain data
-                                    reg_domain_data_for_countrycode(CC);
+            case blockchain:config(poc_version, Ledger) of
+                {ok, V} when V > 10 ->
+                    %% check if the poc 11 vars are active yet
+                    case blockchain_ledger_v1:find_gateway_location(Addr, Ledger) of
```
jeffgrunewald, line 172: this would be a great place for EEP 49 if it's ever officially adopted

Vagabond (approved):

src/miner_lora.erl
vihu: yea, it's not good. Check some of the case ladders in core for maximum annoyance

PR #1167, opened 2021-11-05
Remove poc-v11 check again

vihu:
This re-removes poc-v11 check for getting regulatory domain but keeps specific error handling for when the region vars are not set and also when the regulatory domain is not set.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -166,34 +166,27 @@ reg_domain_data_for_addr(Addr, #state{chain=Chain}) ->
         undefined ->
             {error, no_ledger};
         Ledger ->
-            case blockchain:config(poc_version, Ledger) of
-                {ok, V} when V > 10 ->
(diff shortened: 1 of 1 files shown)


src/miner_lora.erl
```diff
@@ -166,34 +166,27 @@ reg_domain_data_for_addr(Addr, #state{chain=Chain}) ->
...
                                     {error, Reason}
                             end;
+                        {error, region_var_not_set} ->
+                            %% poc-v11 is partially active
+                            lookup_via_country_code(Addr);
+                        {error, regulatory_regions_not_set} ->
+                            %% poc-v11 is partially active
+                            lookup_via_country_code(Addr);
                         {error, Reason} ->
                             {error, Reason}
```
vihu, line 186: Should we do a `lookup_via_country_code` here too?

Vagabond (approved):

vihu:
This was already merged in a separate PR, closing.

PR #1169, opened 2021-11-05
add some basic txn profiling

evanmcc:
get some legwork out of the way for ongoing profiling work.


evanmcc:
this code should be in core

PR #1183, opened 2021-11-09
Discard 0.0 freq witnesses

vihu:
If witnesses are reported with `0.0` frequency, don't send.


vihu:
Closing because this should happen on the poc_statem.

PR #1210, opened 2021-11-16
Add a max block size when constructing blocks

Vagabond:


evanmcc (approved):

src/miner.erl
```diff
@@ -652,12 +652,35 @@ select_transactions(Chain, Txns, BlockCurr, BlockHeightCurr, BlockHeightNext) ->
...
+                                                         {done, Acc};
+                                                     Remainder ->
+                                                         {Remainder, [Txn|Acc]}
+                                                 end
+                                         end, {SizeLimit, []}, ValidTransactions0),
+
+    ValidTransactions = lists:reverse(ValidTransactions1),
+
+    %% any that overflowed go back into the buffer
+    InvalidTransactions = InvalidTransactions1 ++ (ValidTransactions0 -- ValidTransactions),
```
jeffgrunewald, line 683: i assume "invalid" doesn't mean there's anything wrong with the txn data such that it can be successfully validated on later passes through the validation process? otherwise we would need to keep these two ("invalid" txns + valid overflow txns) separate when we pass them back into the buffer?

PR #1212, opened 2021-11-16
Cache einfo in pdict for now

Vagabond:


Vagabond:
Obsoleted by #1213 and friends

PR #1213, opened 2021-11-16
use block_info_v2

andymck:
miner side of https://github.com/helium/blockchain-core/pull/1098


src/miner.erl
```diff
@@ -638,7 +633,7 @@ block_time(Block, Metadata) ->
 -spec select_transactions(
     blockchain:blockchain(),
     blockchain_txn:txns(),
-    blockchain_block:block(),
+    {non_neg_integer(), non_neg_integer()},
```
vihu, line 636:
```suggestion
    ElectionInfo :: {non_neg_integer(), non_neg_integer()},
```

vihu: not needed per se

xandkar: Although they're already named in the func def, so never mind.

evanmcc (approved):

vihu (approved):

PR #1222, opened 2021-11-18
Remove no longer needed GMP, remove gcc from runner image

Vagabond:


Dockerfile
```diff
@@ -13,7 +13,7 @@ ARG EXTRA_BUILD_APK_PACKAGES
 
 RUN apk add --no-cache --update \
     git tar build-base linux-headers autoconf automake libtool pkgconfig \
-    dbus-dev bzip2 bison flex gmp-dev cmake lz4 libsodium-dev openssl-dev \
```
jeffgrunewald, line 16: this seems to cause issues with cargo in my testing:
```
sh: exec: line 1: cargo: not found
```

Dockerfile
```diff
@@ -13,7 +13,7 @@ ARG EXTRA_BUILD_APK_PACKAGES
...
+    dbus-dev bzip2 bison flex cmake lz4 libsodium-dev openssl-dev \
```
jeffgrunewald, line 16:
```suggestion
    dbus-dev bzip2 bison flex gmp-dev cmake lz4 libsodium-dev openssl-dev \
```

Dockerfile
```diff
@@ -42,7 +42,7 @@ FROM ${RUNNER_IMAGE} as runner
 ARG VERSION
 ARG EXTRA_RUNNER_APK_PACKAGES
 
-RUN apk add --no-cache --update ncurses dbus gmp libsodium gcc \
+RUN apk add --no-cache --update ncurses dbus libsodium \
```
Vagabond, line 45:
```suggestion
RUN apk add --no-cache --update ncurses dbus libsodium libgcc \
```

Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BUILDER_IMAGE=erlang:23.3.4.6-alpine
-ARG RUNNER_IMAGE=erlang:23.3.4.6-alpine
+ARG RUNNER_IMAGE=alpine
```
jeffgrunewald, line 2: This is just the default; you’ll need to make sure to update the same environment variable arg in the `.buildkite/scripts/make_image.sh`

Vagabond: maybe you can do that?

Vagabond: I did it.

jeffgrunewald (approved):

lilacanaris2021:
Hi Jeff, I got this message too: 
sh: 1: exec: cargo: not found
===> Hook for compile failed!

Could you please explain what and where I have to change. 

Thanks for your support.  Regards Bert (I`m not an expert)

jeffgrunewald:
@lilacanaris2021 I’m not able to reproduce that issue on either Mac or Linux but it implies you’re not successfully pulling down the rust tool chain in the step before compiling. Since we pull down the rustup from scratch it shouldn’t matter what your host platform is so long as it runs Docker. 

I would check the scroll back and see if there are any silent errors in retrieving the rust tool chain and ensure you’ve got proper dns resolution for that curl. It could just be transient network issue that resolves itself.

lilacanaris2021:
Thanks a lot

PR #1248, opened 2021-11-22
Disable fetching latest snapshot on validators

PaulVMo:
Disable the latest snapshot functionality for syncing validators and instead use only the latest blessed snapshot. Fetching the latest snapshot from a URL creates a security and reliability risk:
1. the hosting URL of the latest snapshot is another attack vector on validators since validators do not have a way to verify the hash of the snapshot
2. not all of the latest snapshots work - as identified in Discord, some of the snapshots cause syncing issues. unlike blessed snapshots, no one reviews or tests the latest snapshots

### DIFF
--- a/config/docker-val.config.src
+++ b/config/docker-val.config.src
@@ -13,6 +13,7 @@
  {blockchain,
   [
    {listen_addresses, ["/ip4/0.0.0.0/tcp/2154"]},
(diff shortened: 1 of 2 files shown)


Vagabond:
This looks reasonable @evanmcc can you check?

evanmcc (approved):

PR #1255, opened 2021-11-24
Updating default OTP versions for dockerized miner/validator builds

jeffgrunewald:
Upgrading dockerized versions of Miner/Validator to default to OTP 24. Also removes libgcc from the build which appears to be unnecessary after comparing to the blockchain-http build.

### DIFF
--- a/.buildkite/scripts/make_image.sh
+++ b/.buildkite/scripts/make_image.sh
@@ -10,12 +10,12 @@ set -euo pipefail
 
 TEST_BUILD=${TEST_BUILD:-0}
 
-ERLANG_IMAGE="23.3.4.7-alpine"
+ERLANG_IMAGE="24-alpine"
 ERLANG_IMAGE_SOURCE="erlang"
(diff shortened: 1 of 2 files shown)


madninja (approved):

Vagabond:
How did you check libgcc was no longer needed? Blockchain-http doesn't have the same dependencies at *all*

jeffgrunewald:
> How did you check libgcc was no longer needed? Blockchain-http doesn't have the same dependencies at _all_

the absence in blockchain-http was just the impetus; i removed it a couple of builds ago and have tested running it _just_ libstdc++ for a couple of builds on this branch now. i can add it back in if you're worried about any hidden footguns but it seems to be fully operational

jeffgrunewald:
any other tests i can run specifically @Vagabond you think might cause this to blow unexpectedly when it's otherwise following the chain?

Vagabond (approved):

PR #1271, opened 2021-11-30
Honor follow_mode set in app env

vihu:
It appears that setting `follow_mode` blockchain app env in dev.config (or other) has no effect because it gets explicitly overwritten in `miner_app.erl` specifically for miner in `gateway` mode.

This fixes it to honor app env and only default it when the app env is `undefined`.

### DIFF
--- a/src/miner_app.erl
+++ b/src/miner_app.erl
@@ -33,15 +33,21 @@ start(_StartType, _StartArgs) ->
     Follow =
         %% validator as the default here because it's a safer failure mode.
         case application:get_env(miner, mode, validator) of
-            %% follow mode defaults to false
+            %% follow mode defaults to false for miner validator mode
             validator -> false;
             gateway ->
-                %% check touchfile, since some people are using it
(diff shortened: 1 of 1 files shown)


src/miner_app.erl
```diff
@@ -33,15 +33,21 @@ start(_StartType, _StartArgs) ->
...
+                case application:get_env(blockchain, follow_mode) of
+                    undefined ->
+                        %% check touchfile, since some people are using it
+                        BaseDir = application:get_env(blockchain, base_dir, "data"),
+                        Filename = filename:join([BaseDir, "validate-mode"]),
+                        case file:read_file_info(Filename) of
+                            {ok, _} -> false;
+                            _ -> true
+                        end;
+                    FollowMode ->
```
abhay, line 48:
```suggestion
                    {ok, FollowMode} ->
```

vihu: good catch, ty!

evanmcc (approved):

PR #1284, opened 2021-12-08
pre-compiling deps to cache and speed up build

jeffgrunewald:
We spend a lot of cycles recompiling dependencies (I'm looking at you, rocksdb). By breaking dependency compilation into a separate image, the docker daemon will be able to take advantage of the image cache when recompiling the project as long as there haven't been any changes to the dependencies in the rebar.config/lock files.

To accomplish this, the build image needs to be broken into two separate builders because the VERSION build arg being part of the image tends to negate the cache immediately. This unfortunately means we _do_ need to duplicate additional environment variables and build arguments that get used by rebar during compilation but this seemed like a small price to pay as I saw it shave a good 20 min of my local build time after the first compilation of a given version of the deps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
 ARG BUILDER_IMAGE=erlang:23.3.4.6-alpine
 ARG RUNNER_IMAGE=alpine
-FROM ${BUILDER_IMAGE} as builder
+FROM ${BUILDER_IMAGE} as deps-compiler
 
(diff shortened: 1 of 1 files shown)


jadeallenx:
Make it so :picard:

jadeallenx:
Need to pull this across to other Docker images we're building too

madninja:
One more rebase @jeffgrunewald

madninja (approved):

PR #1314, opened 2021-12-25
Add initial attempt at miner telemetry

Vagabond:
Merry Christmas

### DIFF
--- a/src/miner_restart_sup.erl
+++ b/src/miner_restart_sup.erl
@@ -94,6 +94,11 @@ init([SigFun, ECDHFun]) ->
     JsonRpcPort = application:get_env(miner, jsonrpc_port, 4467),
     JsonRpcIp = application:get_env(miner, jsonrpc_ip, {127,0,0,1}),
 
+    Telemetry = case application:get_env(miner, telemetry, false) of
+                    true -> [?WORKER(miner_telemetry, [])];
+                    _ -> []
+                end,
+
     ChildSpecs =
         [
          ?WORKER(miner_hbbft_sidecar, []),
@@ -104,7 +109,8 @@ init([SigFun, ECDHFun]) ->
          ] ++
         ValServers ++
         EbusServer ++
-        OnionServer,
+        OnionServer ++
+        Telemetry,
     {ok, {SupFlags, ChildSpecs}}.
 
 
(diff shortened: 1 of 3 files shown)


Vagabond:
The main thing missing here is attestation by including the validator's address and signing it with the validator's key

Vagabond:
Added signatures/identificatioon, but we also need to audit the data sent for sensitive information.

src/miner_restart_sup.erl
```diff
@@ -144,6 +144,12 @@ init(_Opts) ->
                 []
         end,
 
+    Telemetry = case blockchain_utils:get_boolean_os_env_var("SEND_MINER_TELEMETRY", false)
```
madninja, line 147: use the config.src setup instead of this approach?

src/telemetry/miner_telemetry.erl
```diff
@@ -0,0 +1,294 @@
...
+        {Key, Value} -> Value;
+        false -> Default
+    end.
+
+send_telemetry(Data, Signature, Id, Height, Version, URL) ->
+    {ok, {Status, _Body}} = httpc:request(
+            post,
+            {URL,
+                    [
+                        {"X-Telemetry-Signature", ?BIN_TO_B58(Signature)},
```
madninja, line 244: signatures are not addresses.. base64 encode please

src/telemetry/miner_telemetry.erl
```diff
@@ -0,0 +1,294 @@
...
+                        {"X-Telemetry-Miner-Id", Id},
```
madninja, line 245: I see you're also collecting the name but not sending it in a header?

madninja:
so we'll use "telemetry" for this kind of data and "metrics" for the more traditional collection of counts/stats (cc @jeffgrunewald )

src/telemetry/miner_telemetry.erl
jadeallenx: Yeah I wanted a key to use that would be subject to collisions when storing the payloads

src/miner_restart_sup.erl
jadeallenx: OK

jeffgrunewald: i think only the validator profile has config in the src templates at the moment so we’d need to migrate other profiles to that style of config to configure for this telemetry export

jadeallenx: This is primarily intended to be for validators, so I think it's probably ok...

PR #1317, opened 2021-12-27
Fix an issue which was preventing use of RSSIS when available.

ke6jjj:
Fix an issue which was preventing the use of the "rssis" signal report
on a packet, even when it was present and desired. RSSIS or "RSSI of signal"
is an additional packet meta-data item provided by the SX1302 (and other)
packet forwarders, and when present, it more accurately reflects the power
level at which a packet was received. In contrast, the "rssi" value on such
forwarders reflects the "RSSI of channel", which is closer to the RSSI of the
channel noise floor, not the packet that was received.

Since "rssis" is a more accurate gauge of reception power, it is preferred
by PoC v11, but wasn't actually being used because of this issue.

Based on an original commit by Amir Haleem.

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -789,24 +789,37 @@ tmst_to_local_monotonic_time(Tmst_us, PrevTmst_us, PrevMonoTime_us) ->
 %% GWMP JSON V1/V2.
 -spec packet_rssi(map(), boolean()) -> number().
 packet_rssi(Packet, UseRSSIS) ->
-    case maps:get(<<"rssi">>, Packet, undefined) of
-        %% GWMP V2
+    RSSIS = maps:get(<<"rssis">>, Packet, undefined),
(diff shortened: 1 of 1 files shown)


amirhaleem (approved):

src/miner_lora.erl
```diff
@@ -986,3 +999,39 @@ maybe_update_reg_data(#state{pubkey_bin=Addr} = State) ->
...
+        <<"rssis">> => 1,    
+        <<"rssi">> => 2
+    },
+    PacketWithoutRSSIS = #{
+        <<"rssi">> => 2
+    },
+    RSIGPacketWithRSSIS = #{
+        <<"rsig">> => [
+            #{ <<"rssis">> => 1, <<"rssic">> => 2 },
+            #{ <<"rssis">> => 3, <<"rssic">> => 4 },
```
Vagabond, line 1017: Suggest putting the 4 in another entry to show it's maxing on the right value?

Vagabond (approved):

PR #1361, opened 2022-01-14
Add PoC denylist, downloaded from github releases

Vagabond:


src/miner_restart_sup.erl
```diff
@@ -94,6 +94,19 @@ init([SigFun, ECDHFun]) ->
     JsonRpcPort = application:get_env(miner, jsonrpc_port, 4467),
     JsonRpcIp = application:get_env(miner, jsonrpc_ip, {127,0,0,1}),
 
+    DenylistWorker = case application:get_env(miner, denylist_key, undefined) of
```
jeffgrunewald, line 97: could you abstract this out of the supervisor by always adding the denylist worker to the supervision tree and then in the worker's init function return `ignore` if you don't have the right stuffs in the application config?

Vagabond: Oh, I forgot about this, does it work for optional deps?

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,184 @@
...
+            schedule_check(State0);
+        OtherHttpResult ->
+            lager:notice("failed to fetch github release info ~p", [OtherHttpResult]),
+            schedule_check(State0)
+    end;
+handle_info(Msg, State) ->
+    lager:info("unhandled info msg ~p", [Msg]),
+    {noreply, State}.
+
+handle_cast(Msg, State) ->
```
jeffgrunewald, line 168: you don't actually need this right, unless you just want to silence the default "unhandled message warning log"?

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,184 @@
...
+    lager:info("unhandled cast msg ~p", [Msg]),
+    {noreply, State}.
+
+handle_call(Msg, _From, State) ->
```
jeffgrunewald, line 172: looks like you haven't handled the two calls in the api yet

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,184 @@
...
+                                                                            case file:rename(TmpDenyFile, DenyFile) of
+                                                                                ok ->
+                                                                                    ok;
+                                                                                {error, RenameReason} ->
+                                                                                    lager:notice("failed to rename ~p to ~p: ~p", [TmpDenyFile, DenyFile, RenameReason])
+                                                                            end;
+                                                                        {error, WriteReason} ->
+                                                                            lager:notice("failed to write denyfile ~p to disk ~p", [TmpDenyFile, WriteReason])
+                                                                    end,
+                                                                    ok = persistent_term:put(?MODULE, Filter),
```
jeffgrunewald, line 134: seems like a little wrangling with that tmpfile on the disk-loaded file and you could combine this with the read-file/load persistent term functionality in the `init/1`

src/poc/miner_poc_denylist.erl
Vagabond: No, not yet, was going to do that later, maybe, so we can gossip this file around.

src/poc/miner_poc_denylist.erl
Vagabond: Is this true on OTP22 as well?

src/miner_restart_sup.erl
jeffgrunewald: like if we make the `xorf` app optional the way we've been talking about optional deps related to the tpm module branch? not sure but i've used this previously for enabling/disabling optional features in a supervision tree. if you're worried an optional dep could be missing could you make the init function do something like this:
```
case {application:get_env(miner,` denylist_config, undefined), code:ensure_loaded(xorf)} of
    {undefined, _} ->
        ignore;
    {Config, {module, xorf}} ->
        rest_of_the_worker_init,
        {ok, schedule_check(State, 0)
end
```

src/poc/miner_poc_denylist.erl
jeffgrunewald: i guess nvm; i can't find any evidence that this functionality exists at _all_ for `handle_call/3` or `handle_cast/2`; as of OTP 20 `handle_info, terminate, and code_change` are optional but those are the only ones.

Vagabond: yeah, I knew those were optional

src/miner_restart_sup.erl
Vagabond: it doesn't need to be an optional app, we just need to allow the denylist to not be configured

jeffgrunewald: Then it should work to make the worker accept an empty list in the supervisor childspec and then grab those config values in its init function or else return ignore if they’re not present. I did something similar previously where I checked if the child supervisor was `enabled` and if it was false I returned ignore and exited the process cleanly and in those cases the rest of that process’s config was omitted from the app env completely without any problems.

src/poc/miner_poc_statem.erl
```diff
@@ -108,7 +108,12 @@ receipt(Address, Data, PeerAddr) ->
     gen_statem:cast(?SERVER, {receipt, Address, Data, PeerAddr}).
 
 witness(Address, Data) ->
-    gen_statem:cast(?SERVER, {witness, Address, Data}).
+    case miner_poc_denylist:check(Address) of
+        true ->
+            ok;
```
abhay, line 113:
```suggestion
            lager:info("denylist: dropping witness from ~p", [Address]),
            ok;
```

abhay, line 113: add some kind of explicit logging here?

src/poc/miner_poc_statem.erl
```diff
@@ -554,17 +559,22 @@ handle_challenging({Entropy, TargetRandState}, Target, Gateways, Height, Ledger,
...
-                    {next_state, receiving, save_data(Data#data{state=receiving, challengees=lists:zip(Path, LayerData),
-                        packet_hashes=lists:zip(Path, LayerHashes)})};
-
-                {error, Reason} ->
-                    lager:error("failed to dial 1st hotspot (~p): ~p", [P2P, Reason]),
-                    lager:info("selecting new target"),
-                    handle_targeting(Entropy, Height, Ledger, Data#data{retry=Retry-1})
+            case lists:any(fun miner_poc_denylist:check/1, Path) of
+                true ->
+                    {next_state, requesting, save_data(Data#data{state=requesting})};
```
abhay, line 564: add explicit logging here that the challenge is being dropped due to denylist?

src/poc/miner_poc_denylist.erl
Vagabond: This is tricky because the error paths differ, so I didn't do it.

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,200 @@
...
+                                    0
+                            end;
+                        false ->
+                            0
+                    end,
+    {ok, schedule_check(#state{type=Type, url=URL, key=Key, version=FilterVersion}, 0)}.
+
+handle_info(check, #state{type=github_release, url=URL, key=Key, version=Version, etag=Etag}=State) ->
+    %% pull the release definition
+    case httpc:request(get, {URL, [{"user-agent", "https://github.com/helium/miner"}] ++ [ {"if-none-match", Etag} || Etag /= undefined] }, [], [{body_format, binary}]) of
```
jadeallenx, line 91: Does this Etag need quotes to work like on S3?

jadeallenx (commented):
It feels sort of like we might want to have an http client ... since we seem to keep adding HTTP calls to fetch data. Seems silly to keep writing the same logic over and over...

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,200 @@
+-module(miner_poc_denylist).
+
+-behaviour(gen_server).
+
+-record(state, {
+          type,
```
vihu, line 6:
```suggestion
          type = github_release :: atom()
```

vihu, line 6:
```suggestion
          type = github_release :: atom(),
```

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,200 @@
...
+          key,
```
vihu, line 7:
```suggestion
          key = undefined :: libp2p_crypto:pubkey_bin(),
```

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,200 @@
...
+          url,
```
vihu, line 8:
```suggestion
          url = undefined :: string(),
```

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,200 @@
...
+          version,
```
vihu, line 9:
```suggestion
          version :: pos_integer(),
```

src/poc/miner_poc_denylist.erl
```diff
@@ -0,0 +1,200 @@
...
+          etag
```
vihu, line 10:
```suggestion
          etag :: string(),
```

vihu, line 10:
```suggestion
          etag :: string()
```

src/poc/miner_poc_denylist.erl
Vagabond: apparently not

evanmcc (approved):

PR #1368, opened 2022-01-19
wiring hotfix patch directory into miner code path

jeffgrunewald:
In order to simplify hot-patching containerized miners, we need to include a directory at the head of the application's code path and include it in the build container, while gracefully degrading the patch feature when not running in an environment with this setup.

Introduces the new code directory addition at the head of the path, configurable in the sys.config file(s) and adds some volume declarations in the container metadata that documents the mount points where bind mounts are expected.

note: also renames the `miner-testnet.config` file to `docker-testnet.config` to better align with the other config files based on their usage, given this file is only used when the rebar release target is the dockerized miner on the tesnet


jadeallenx (approved):

src/miner_app.erl
```diff
@@ -30,6 +30,20 @@ start(_StartType, _StartArgs) ->
...
     application:set_env(rocksdb, global_opts, Opts),
 
+    HotfixDir = application:get_env(miner, hotfix_dir, "hotfix"),
+
+    case filelib:is_dir(HotfixDir) of
+        true ->
+            case code:add_patha(HotfixDir) of
+                true ->
+                    lager:info("added ~p to the code path", [HotfixDir]);
+                {error, bad_dir} ->
```
JayKickliter, line 40: ```The pattern {'error', 'bad_dir'} can never match the type {'error','bad_directory'}```
```suggestion
                {error, bad_directory} ->
```

JayKickliter (approved):

src/miner_app.erl
```diff
@@ -30,6 +30,20 @@ start(_StartType, _StartArgs) ->
                       BBOpts ++ [{block_cache, Cache}]}],
     application:set_env(rocksdb, global_opts, Opts),
 
+    HotfixDir = application:get_env(miner, hotfix_dir, "hotfix"),
```
vihu, line 33: Should the default be `"/opt/miner/hotfix"` here?

vihu (approved):
1 comment, looks good otherwise.

src/miner_app.erl
JayKickliter: I'm not sure there needs to be a default.

jeffgrunewald: in any case i was trying to follow the pattern we have for `BaseDir` where we default the value to `"data"` and then in most of the configs and run scripts override it to `"/var/data"`

vihu: > in any case i was trying to follow the pattern we have for `BaseDir` where we default the value to `"data"` and then in most of the configs and run scripts override it to `"/var/data"`

I think that makes sense so in a dev/local env the `hotfix` dir will be just next to data presumably. That's probably okay.

PR #1370, opened 2022-01-19
Implement ECC API through rust gateway via GRPC

jeffgrunewald:
Bakes the rust gateway into the erlang miner as a dependent external service, managed by an erlang port started up at the beginning of the application.

- Implements the gateway port management process that monitors the external gateway-rs process and controls its lifecycle
- Implements the gateway_ecc_worker that controls interactions with the rust gateway process over grpc
- Reworks the startup process to ensure the gateway port and the grpc ecc worker are started and ready to feed the necessary crypto functions into the downstream processes
- Make functionality to pull down and build the grpc autogen (gpb/grpc_client) code as well as the gateway-rs repo and build it as part of the miner rebar compilation process


Makefile
```diff
@@ -60,3 +74,26 @@ devrel:
 devrelease:
 	$(REBAR) as dev release
 
+grpc:
+	@echo "generating miner grpc services"
+	REBAR_CONFIG="config/grpc_client_gen.config" $(REBAR) grpc gen
```
andymck, line 79: prob a good idea to rename this based on the service it supports, something like grpc_client_gen_local.config

I will do the same on the POC branch for the gateway service.  that way we avoid any conflict,  allows each service to have its own config file with its own gpb options etc

src/miner_gateway_ecc_worker.erl
```diff
@@ -0,0 +1,148 @@
...
+    gen_server:cast(?MODULE, reconnect).
+
+start_link(Options) when is_list(Options) ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [Options], []).
+
+init([Options]) ->
+    Transport = proplists:get_value(transport, Options, tcp),
+    Host = proplists:get_value(host, Options, "localhost"),
+    Port = proplists:get_value(port, Options, 4468),
+    {ok, Connection} = grpc_connect(Transport, Host, Port),
```
andymck, line 64: maybe push this out of the init.  grpc_connect has reconnect support with 10 seconds between retries.  Handling this in init could potentially hang your supervisor startup sequence whilst it waits for grpc connect to complete

Makefile
jeffgrunewald: oof; i had just figured they'd merge together but yeah, if you need different gpb options that will get messy; will do.

src/miner_gateway_ecc_worker.erl
jeffgrunewald: since this is now our only point of access to the pubkey which we need to pass to the downstream supervisors (blockchain_sup in critical path at least) and also those signing and ecdh functions, i would think we would _want_ to explicitly hang in the supervisor init block here in order to ensure the other supervisors have the info they need. if we somehow didn't connect immediately and had to wait 10 seconds in order to retrieve the pubkey i think the miner would panic

andymck: k, i hadnt noted that and makes sense.  I will follow through the sup sequence again

Makefile
```diff
@@ -9,15 +9,35 @@ else
   CIBRANCH=$(shell git rev-parse --abbrev-ref HEAD | sed 's/\//-/')
 endif
 
+ifndef RELEASE_ARCH
+override RELEASE_ARCH=amd64
+endif
+
+ifndef LIBC
+override LIBC=musl
+endif
```
JayKickliter, line 18: My gut tells me these shouldn't be in the Makefile at all. But if they must be, `?=` is the idiomatic way of conditionally setting variables.
```suggestion
RELEASE_ARCH?=amd64
LIBC?=musl
```

Makefile
```diff
@@ -9,15 +9,35 @@ else
...
+
+ifeq ($(RELEASE_ARCH), amd64)
+  RUST_TARGET=x86_64-unknown-linux-$(LIBC)
+else
+  RUST_TARGET=aarch64-unknown-linux-$(LIBC)
+endif
```
JayKickliter, line 24: This breaks local builds for me on macOS

Makefile
jeffgrunewald: that switch was debugging build issues that turned out to be a mismatch between the host and the container during build time but I left them b/c supposedly the `override` syntax will handle when the value is defined but `null` whereas `?=` will only handle when the value isn't yet defined? in any case, i'm going to try and strip those out and default the target to "whatever you're building for" but i need to figure out what that behavior looks like inside a production build alpine container and make sure i can detect and override that if needed.

Vagabond:
One thing to make sure of here: Existing miner configs should continue to work with these changes. As in the "old" way to configure disk keys or ECC should be honored without changes. Also strongly recommend not requiring docker changes so this can be a drop-in replacement upgrade.

Vagabond:
Otherwise, if absolutely necessary, very clear upgrade instructions are required.

config/docker-testnet.config.src
```diff
@@ -23,6 +23,7 @@
      {network, testnet},
      {api_base_url, "https://testnet-api.helium.wtf/v1"},
      {jsonrpc_ip, {0,0,0,0}}, %% bind jsonrpc to host when in docker container
+     {gateway_and_mux_enable, ${GW_AND_MUX}},
```
Vagabond, line 26: I think adding these ENV vars (ones that cause the sys.config to become invalid if they're not supplied) is probably a bad idea, or at least one that should be re-examined. At a minimum putting quotes around the vars would mean, at worst, that the var would be set to the empty string?

src/miner_critical_sup.erl
```diff
@@ -86,10 +94,18 @@ init([PublicKey, SigFun, ECDHFun, ECCWorker]) ->
...
-        ] ++
-        ConsensusMgr,
+    ChildSpecs0 = [?SUP(blockchain_sup, [BlockchainOpts])] ++ ConsensusMgr,
+    GatewayAndMux = case application:get_env(miner, gateway_and_mux_enable) of
+                        {ok, true} -> true;
+                        _ -> false
+                    end,
+    ChildSpecs = case {GatewayAndMux, application:get_env(blockchain, key)} of
+                     {false, {ok, {ecc, _}}} ->
+                         [
```
Vagabond, line 104:
```suggestion
                         % Miner retains control of key access
                         [
```

config/test.config
```diff
@@ -42,7 +40,7 @@
    {election_interval, 10},
    {dkg_stop_timeout, 15000},
    {write_failed_txns, true},
-   {radio_device, undefined},
+   {radio_device, {{127, 0, 0, 1}, 1681, {127, 0, 0, 1}, 31341}},
```
Vagabond, line 43: Why did this change?

src/miner_gateway_ecc_worker.erl
```diff
@@ -0,0 +1,159 @@
...
+
+%% Pass an ecc public key to the rust gateway and return a point on the
+%% eliptic curve as a binary
+-spec ecdh(libp2p_crypto:pubkey()) -> {ok, Preseed :: binary()} | {error, term()}.
+ecdh({ecc_compact, _Bin} = PubKey) ->
+    gen_server:call(?MODULE, {ecdh, PubKey}, ?CALL_TIMEOUT).
+
+%% Trigger a reconnect of the grpc_client connection
+-spec reconnect() -> ok.
+reconnect() ->
```
Vagabond, line 59: When is reconnect needed?

src/miner_gateway_port.erl
```diff
@@ -0,0 +1,148 @@
...
+    {noreply, State}.
+
+handle_info({Port, {exit_status, Status}}, #state{port = Port} = State) ->
+    lager:warning("gateway-rs process ~p exited with status ~p, restarting", [Port, Status]),
+    ok = cleanup_port(State),
+    NewState = open_gateway_port(State#state.keypair, State#state.transport, State#state.host, State#state.tcp_port),
+    ok = miner_gateway_ecc_worker:reconnect(),
+    {noreply, NewState};
+handle_info({Port, {data, LogMsg}}, #state{port = Port} = State) ->
+    Lines = binary:split(LogMsg, <<"\n">>, [global]),
```
Vagabond, line 60: If the last line is incomplete doesn't this mean you may log a partial line?

src/miner_gateway_port.erl
```diff
@@ -0,0 +1,148 @@
...
+            lager:debug("[ gateway-rs ] ~s", [Statement]);
+        <<" DEBUG ", Statement/binary>> ->
+            lager:debug("[ gateway-rs ] ~s", [Statement]);
+        <<" INFO ", Statement/binary>> ->
+            lager:info("[ gateway-rs ] ~s", [Statement]);
+        <<" WARN ", Statement/binary>> ->
+            lager:warning("[ gateway-rs ] ~s", [Statement]);
+        <<" ERROR ", Statement/binary>> ->
+            lager:error("[ gateway-rs ] ~s", [Statement]);
+        _ -> lager:debug("unhandled info ~p", [Line])
```
Vagabond, line 135: Arguably this is an exceptional condition and should be logged more obviously?

src/miner_keys.erl
```diff
@@ -128,27 +198,68 @@ key_config() ->
...
     lists:foreach(fun(Term) -> io:format("~tp.~n", [Term]) end, Props),
     rpc_ok.
 
+key_proplist_to_uri(Props) ->
+    Bus = proplists:get_value(bus, Props, "i2c-1"),
+    Address = proplists:get_value(address, Props, 16#60),
+    KeySlot = proplists:get_value(key_slot, Props, 0),
+    Network = application:get_env(blockchain, network, mainnet),
+    PortAndQuery = lists:flatten(io_lib:format(":~p?slot=~p&network=~s", [Address, KeySlot, Network])),
+    "ecc://" ++ Bus ++ PortAndQuery.
```
Vagabond, line 241: Why not do this in the format string on the previous line?

src/miner_keys.erl
```diff
@@ -128,27 +198,68 @@ key_config() ->
...
+                               ({"network", Network}, Acc) ->
+                                   [{network, Network} | Acc];
+                               (_, Acc) -> Acc
+                            end, [], uri_string:dissect_query(Query)),
+    [{bus, Bus}, {address, Address}] ++ QueryList.
 %%
 %% Utilities
 %%
 
+key_proplist_to_map(Options) when is_list(Options) ->
```
Vagabond, line 256: You may want to match that this is a list of tuples, not a string?

Makefile
```diff
@@ -60,3 +69,49 @@ devrel:
...
+
+clean_grpc:
+	@echo "cleaning miner grpc services"
+	rm -rf $(GRPC_SERVICES_DIR)
+
+external_svcs:
+	@echo "cloning external dependency projects"
+	@echo "--- gateway-rs ---"
+	$(call clone_project,gateway-rs,$(GATEWAY_RS_VSN))
+	cd ./external/gateway-rs && cargo build --release && cd ../../
```
JayKickliter, line 89:
```suggestion
	(cd ./external/gateway-rs && cargo build --release)
```

src/miner_mux_port.erl
```diff
@@ -0,0 +1,124 @@
+%%%-------------------------------------------------------------------
+%% @doc miner gwmp-mux service manager
+%% Start and monitor the port that runs the external rust-based semtech
+%% gwmp mux service
+%% @end
+%%%-------------------------------------------------------------------
+-module(miner_mux_port).
```
Vagabond, line 7: Can we dedup this with the other port module any?

src/miner_keys.erl
jeffgrunewald: i think i was probably using `~p` there and getting the extra escaped string quotes in the result; forgot to collapse these when i added the network which is also either a string or an atom

src/miner_keys.erl
jeffgrunewald: like `key_proplist_to_map([{_, _} | _] = Options) ->` or is there a different guard i'm missing?

src/miner_port_services_sup.erl
```diff
@@ -0,0 +1,99 @@
...
+            MuxOpts = [
+                {host_port, application:get_env(miner, mux_host_port, 1680)},
+                {client_ports, application:get_env(miner, mux_client_ports, [1681, 1682])}
+            ],
+
+            ChildSpecs =
+                [
+                    ?WORKER(miner_gateway_port, [GatewayPortOpts]),
+                    ?WORKER(miner_gateway_ecc_worker, [GatewayECCWorkerOpts]),
+                    ?WORKER(miner_mux_port, [MuxOpts])
```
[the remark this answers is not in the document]
jeffgrunewald: i put the mux last so the gateway was up and running before the mux was able to send it any packets, the alternative being if the mux were first it could be routing packets to the gateway that weren't being handled yet while it was still starting.
if the mux dies it should get restarted by the supervisor and in the meantime the other two just aren't receiving any packets so they're just waiting as if the mux were running all along and not receiving packets

src/miner_mux_port.erl
jeffgrunewald: like a port behaviour? are these likely to stick around long enough to abstract that commonality out or do we want to just maintain until we transition to lgws?

config/test.config
jeffgrunewald: testing cruft; reverted

src/miner_gateway_port.erl
jeffgrunewald: testing this out i didn't see any weird behavior if the final complete log message _didn't_ include a trailing `\n` so the final item in the split list should be as complete or incomplete as whatever the gateway sends over stdout; unless i stuffed any incomplete lines into the state and waited to see if they were completed how would i ensure i incomplete lines weren't logged? or are you thinking it'd be better to _not_ to log incomplete lines at all?

config/docker-testnet.config
```diff
@@ -23,8 +23,9 @@
      {network, testnet},
      {api_base_url, "https://testnet-api.helium.wtf/v1"},
      {jsonrpc_ip, {0,0,0,0}}, %% bind jsonrpc to host when in docker container
+     {gateway_and_mux_enable, true},
      {use_ebus, false},
-      {radio_device, { {0,0,0,0}, 1680,
+      {radio_device, { {0,0,0,0}, 1681,
```
jeffgrunewald, line 28:
```suggestion
      {radio_device, { {0,0,0,0}, 1680,
```

JayKickliter (approved):

PR #1382, opened 2022-01-22
revert otp24

evanmcc:
looks like this is maybe causing some issues, revert to the older version for now.


.buildkite/scripts/make_image.sh
```diff
@@ -12,7 +12,7 @@ set -euo pipefail
 
 TEST_BUILD=${TEST_BUILD:-0}
 
-ERLANG_IMAGE="24-alpine"
+ERLANG_IMAGE="23.3.4.7-alpine"
```
jeffgrunewald, line 15: probably not an issue (especially if we're sure this is a stable release) but the latest 23 is 23.3.4.10 fwiw; `23-alpine` or anything in between will also suffice to float to the latest patch/point release

jeffgrunewald (approved):

evanmcc:
closing for now, not clear that this was warranted.

PR #1399, opened 2022-01-28
Order HBBFT txn buffer by fee value

Vagabond:


andymck (approved):
lgtm, need to sanity test with final version of hbbft changes but miner side looks fine

src/miner_hbbft_sidecar.erl
```diff
@@ -146,6 +146,8 @@ handle_call({submit, Txn}, From,
                             case blockchain_txn:absorb(Txn, Chain) of
                                 ok ->
                                     spawn(fun() ->
+                                                  %% this will now return {ok, Position, Length}
```
abhay, line 149: or just plain `ok` until the TODO in the handler is implemented? 

Separately, there are a few other cases in the sidecar where this is called which may be worth documenting as well.

andymck: it it helps, i will be changing sidecar in another PR to use the responses....

src/handlers/miner_hbbft_handler.erl
```diff
@@ -214,15 +214,22 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
...
             {reply, ok, ignore};
         false ->
-            case hbbft:input(State#state.hbbft, blockchain_txn:serialize(Txn)) of
-                {NewHBBFT, ok} ->
-                    {reply, ok, [], State#state{hbbft=NewHBBFT}};
+            Fee = blockchain_txn:fee(Txn),
+            Comparator = fun(OtherSerializedTxn) ->
+                                 OtherTxn = blockchain_txn:deserialize(OtherSerializedTxn),
+                                 %% we want to insert before any txn with a lower fee
+                                 blockchain_txn:fee(OtherTxn) < Fee
```
Vagabond, line 224:
```suggestion
                                 Fee > blockchain_txn:fee(OtherTxn)
```

src/handlers/miner_hbbft_handler.erl
```diff
@@ -214,15 +214,22 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
     Buf = hbbft:buf(HBBFT),
     case lists:member(blockchain_txn:serialize(Txn), Buf) of
         true ->
+            %% TODO return the existant position in Buf
             {reply, ok, ignore};
         false ->
-            case hbbft:input(State#state.hbbft, blockchain_txn:serialize(Txn)) of
-                {NewHBBFT, ok} ->
-                    {reply, ok, [], State#state{hbbft=NewHBBFT}};
+            Fee = blockchain_txn:fee(Txn),
```
Vagabond, line 220: I think we should give state channel closes a `virtual` fee here so they sort before poc txns, so router can get them closed more easily

evanmcc (approved):
this seems reasonable once hbbft is merged and updated in the rebar config and lockfile

evanmcc (approved):

Issue #1385, opened 2022-01-24
validator deb package doesn't include a genesis block


kellybyrd:
The deb package doesn't include a genesis block. To make the package more useful for average users, this package should include a genesis block that is auto-loaded, mainnet for main builds and testnet for test builds.

meowshka:
Genesis block was added, closing. 
https://github.com/helium/miner/pull/1402/commits/03cb7bfc85e5cd5cfd9fc7b7633378dea397090c

Issue #1061, opened 2021-09-15
Remove dependence on hardcoded user in DEB package


PaulVMo:
The DEB package version of the miner has a hardcoded user name of 'ubuntu' in the systemd service definition. This will cause it to fail to start on non-Ubuntu OS's and on Ubuntu machines where the user has changed the default user name.

https://github.com/helium/miner/blob/838fe86d3239f8e9d2559160a88b279719475e4b/deb/miner.service#L9

kellybyrd:
IMO the deb package should create a dedicated user and group, then have the service file run as that user. I suggest `helium:helium`

PR #1402, opened 2022-01-28
deb package enhancements

PaulVMo:
The deb package provides an alternative to building from source or running as docker. This PR attempts to improve the usability of the package for validator operators with the following changes:

1. Changes the file install location to `/opt/miner` to match defaults in sys.config and enable adding genesis block
2. Includes the genesis file in the package at `/opt/miner/update` so it automatically gets loaded. Selects network based on BUILD_NET arg, default=mainnet (see also issue #1385 )
3. Enables and runs the systemd service after install
4. Changes user and group to `helium` for running the miner and file ownership. User and group are created as part of before-install script. Previously the miner was run as `ubuntu` which may not exists for all operators. (issue #1061 )
5. Adds symlink in `/usr/local/bin/` to the miner executable for ease of use running miner commands
6. Updates the target distro in the packagecloud repo from Ubuntu 18.04 LTS to 20.04 LTS

Thanks to @kellybyrd and @ericjohncarlson for support on these changes.

### DIFF
--- a/deb/after_remove.sh
+++ b/deb/after_remove.sh
@@ -1,0 +1,1 @@
+rm -f /usr/local/bin/miner
--- a/.buildkite/scripts/make_deb.sh
+++ b/.buildkite/scripts/make_deb.sh
@@ -2,8 +2,13 @@
 
 set -euo pipefail
 
+BUILD_NET="${BUILD_NET:-mainnet}"
+
+
 VERSION=$(echo $VERSION_TAG | sed -e 's,validator,,')
 DIAGNOSTIC=1 ./rebar3 as validator release -n miner -v ${VERSION} || ./rebar3 as validator release -n miner -v ${VERSION}
+
+wget -O /tmp/genesis https://snapshots.helium.wtf/genesis.${BUILD_NET}
 
 fpm -n validator \
     -v "${VERSION}" \
@@ -11,6 +16,22 @@ fpm -n validator \
     -t deb \
     --depends libssl1.1 \
     --depends libsodium23 \
+    --depends libncurses5 \
+    --depends dbus \
+    --depends libstdc++6 \
     --deb-systemd deb/miner.service \
+    --before-install deb/before_install.sh \
+    --after-install deb/after_install.sh \
+    --after-remove deb/after_remove.sh \
+    --before-upgrade deb/before_upgrade.sh \
+    --after-upgrade deb/after_upgrade.sh \
     --deb-no-default-config-files \
-    _build/validator/rel/=/var/helium
+    --deb-systemd-enable \
+    --deb-systemd-auto-start \
+    --deb-systemd-restart-after-upgrade \
+    --deb-user helium \
+    --deb-group helium \
+    _build/validator/rel/=/opt \
+    /tmp/genesis=/opt/miner/update/genesis
+
+rm -f /tmp/genesis
--- a/.buildkite/scripts/packagecloud_upload.sh
+++ b/.buildkite/scripts/packagecloud_upload.sh
@@ -9,6 +9,6 @@ PKGNAME="validator_${TAG}_amd64.deb"
 buildkite-agent artifact download ${PKGNAME} .
 
 curl -u "${PACKAGECLOUD_API_KEY}:" \
-     -F "package[distro_version_id]=190" \
+     -F "package[distro_version_id]=210" \
      -F "package[package_file]=@${PKGNAME}" \
      https://packagecloud.io/api/v1/repos/helium/validators/packages.json
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ data
 \#*
 src/pb
 .repos/
+*.deb
 
 # make build output for external deps
 external/
--- a/deb/after_install.sh
+++ b/deb/after_install.sh
@@ -1,0 +1,13 @@
+# add miner to /usr/local/bin it appears in path
+ln -s /opt/miner/bin/miner /usr/local/bin/miner
+
+# if upgrading from old version with different file location, move miner date files to the new location
+if [ -e /var/helium/miner/data/miner/swarm_key ] && [ ! -e /opt/miner/data/miner/swarm_key ]; then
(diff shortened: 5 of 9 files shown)


ericjohncarlson:
Looks great!  Well-needed updates to this package.  This really should be a primary method of validator installation for most operators.

deb/after_remove.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+rm /usr/local/bin/miner
```
kellybyrd, line 3: nit: maybe do `rm -f` here just to be sure it isn't interactive? Probably not needed.

kellybyrd (commented):
Overall, I think `/opt/helium/validator/` makes more sense for the base dir, or at least `/opt/validator/` but if that doesn't bother anyone else, that's fine.

deb/after_remove.sh
PaulVMo: good catch. updated.

PaulVMo:
> Overall, I think `/opt/helium/validator/` makes more sense for the base dir, or at least `/opt/validator/` but if that doesn't bother anyone else, that's fine.

Thanks for the feedback. I went with `/opt/miner` because the existing code looks in `/opt/miner/update` for a genesis block. This is set in the sys.config file and I did not want to have to change that which would impact the docker image or add another sys.config that would need to be also maintained.

Vagabond:
This sounds great, but I'll leave it to the people who built the packaging in the first place to confirm.

I do think we should keep the miner data on /var but move the code to /opt though.

kellybyrd:
> I do think we should keep the miner data on /var but move the code to /opt though.
I would love this too, but Paul and others were talking about this, and our thinking was that current validator operators are either:
- built source where `data` defaults to next to the `bin` directory
- docker where there is just a mapped data dir.
- some crazy FreeBSD folks doing very custom stuff :-)

The thinking was this "all in one directory" would be the least surprising to most people and the smallest change for a PR. Since the one package contains binaries, config, and data, /opt was the best place overall. 

But, ya I would love to see more traditional locations for binaries, config, and data.

PaulVMo:
I agree that `/var` is a better location for the data. Anyone have thoughts on how to make that happen? The validator default val.config.src puts the data in `data/` which is relative to the base code directory (/opt/miner).

I tried setting the working directory to `/var/helium/miner` in the systemd .service file however the miner still executes with the pwd of `/opt/miner`. Another approach could be modifying sys.config.src as part of the after-install script but that feels too hacky and harder to maintain.

The cleanest option that I see is to add another sys.config.src variant (e.g., deb-val.config.src) and add another profile in rebar.config for deb_val. This is yet another config variant that will need to be maintained if that is okay. Thoughts?

kellybyrd:
A couple of other options:
* A `DATA_DIR` env var that works like the others, that gets templated into sys.config by the `bin/miner` shell script.
* Some way to templatize `config.src` as part of the deb build, the build itself would have to know it is a deb target vs a tar or src target, maybe this could just be awk in `make_deb.sh`, similar to how it is done in `bin/miner` 

If we're separating out data and binaries, what about logs? those properly go in /var/log/<something>/, but I expect that would surprise quite a few validator operators, and would definitely require tweaking existing aliases and metrics gathering. Maybe we consider each of these (data and log) as incremental improvements over this. This PR just makes the deb much more usable, those would make it fit more like a proper package.

PaulVMo:
I have updated the deb package to put the code at `/opt/miner`, logs at `/var/log/miner` and the blockchain data at `/var/data/miner`. I look forward to any other feedback.

kellybyrd:
Nice updates @PaulVMo ! This looks like a real package now!

ericjohncarlson:
A lot of hard work here for the benefit of most validator operators and the Helium network.  Thanks @PaulVMo.

PaulVMo:
I updated the PR to revert the change to split the location of code, data and logs. For now, I have it all going to /opt/miner. I am at a loss on how to retain the existing val.src.config and also have the deb package put the code, logs and data in different directories. And, in order to get automatic loading of the genesis block, it must be in /opt/miner/update per the sys.config file.

I tried environment variable replacement in val.src.config (similar to how VALIDATION_WIDTH works). However, if the environment var is not set, then the config var is read as the empty string "". I was hoping that the default in application:get_env/3 call in the code would get picked up. However, it instead read "" from the file. Code changes might be possible to interpret "" and use a different default although lager hasn't it own default of undefined.

Please let me know if there are other thoughts on how to approach this - otherwise, I recommend proceeding with this as is in order to get the benefit of the genesis block.

jeffgrunewald (approved):

jontow:
We were throwing around some ideas internally on how to safely migrate users of the existing deb pkg so swarm_key and friends don't change or get removed.  I came up with the following concept that seemed an agreeable way to start a discussion.

It seems as though we need something in deb/before_install.sh to facilitate this, and this is what was in my head:
```
if [ -e /var/data/miner/miner/swarm_key ] && [ ! -e /opt/miner/miner/swarm_key ]; then
    mkdir -p /opt/miner && cp -Rp /var/data/miner/miner /opt/miner/miner
fi
```

This is untested, but aims to do things predictably:
* leaves old location in place, untouched (confusing, but safe)
* only copies if old location exists
* does not replace if new key exists for whatever reason (safe, predictable)

OrmEmbaar:
@jontow Would it maintain safety but reduce confusion if we rename the old `swarm_key` to `swarm_key.old` or similar? Just as a signpost to users that they're not looking at a file which is in active use.

PaulVMo:
Pushed some updates and believe this is good to go now. I added before install/upgrade functionality to move miner data directory to /opt/miner if it exists in a different location. I tested it by installing this version as an upgrade to the existing packagecloud version. The data was moved out of /var/helium/miner and the validator picked up at the same height and continued syncing.

I also changed the distro ID in the packagecloud upload script to Ubuntu 20.04. Note, for future research, it appears that there is an option to create a repo on package cloud as distro/version of any/any which the Nova team may want to investigate.

Issue #1409, opened 2022-01-31
GA tagged images not being made on releases?


Jonzky:
Not too sure if it's intentional or a bug, but it looks like GA tagged images are not being pushed alongside the :latest tag.

It looks like that part of the _beta / _alpha addition the GA tagging was removed.
https://github.com/helium/miner/commit/2adf5e8a28c73f335f782a1227a8ef725560ec3d

Just wanted to clarify if this was intentional.

Some manufacturers (or Nebra at least) have not released 2022.01.27 as they are waiting and reliant on the GA.
https://github.com/NebraLtd/hm-miner/pull/72

Vagabond:
Does line 77 not check for the GA tag?

Jonzky:
@Vagabond It does, but it doesn't tag the container image - it effectively strips it and only pushed an image tagged with :latest.

This isn't an issue for me, but it seems like Nebra is holding out on the release until there is a GA image.

Jonzky:
Big thanks for helping with this and getting it resolved so fast. Closing the issue as it should be fixed.

PR #1410, opened 2022-01-31
preserving and pushing _GA-suffixed tags

jeffgrunewald:
Ensure docker tag builds triggered by a `_GA` git tag carry into the docker artifact pushed to the quay registry for makers that trigger deployments and/or releases based on explicit GA'd semvers instead of a floating `latest` tag release.

Addresses issue #1409

### DIFF
--- a/.buildkite/scripts/make_image.sh
+++ b/.buildkite/scripts/make_image.sh
@@ -86,11 +86,16 @@ if [[ "$BUILD_TYPE" == "miner" ]]; then
         echo "Miner tag update detected; Updating ${UPDATE_TAG} on ${REGISTRY_HOST}..."
 
         UPDATE_TAG="${UPDATE_TAG}-${IMAGE_ARCH}"
-        DOCKER_NAME=$(echo "$DOCKER_NAME" | sed -e 's/_GA//' -e 's/_alpha//' -e 's/_beta//')
+        SEMVER_TAG=$(echo "$DOCKER_NAME" | sed -e 's/_GA//' -e 's/_alpha//' -e 's/_beta//')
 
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/make_image.sh
```diff
@@ -86,12 +86,17 @@ if [[ "$BUILD_TYPE" == "miner" ]]; then
...
-        DOCKER_NAME=$(echo "$DOCKER_NAME" | sed -e 's/_GA//' -e 's/_alpha//' -e 's/_beta//')
+        SEMVER_TAG=$(echo "$DOCKER_NAME" | sed -e 's/_GA//' -e 's/_alpha//' -e 's/_beta//')
 
-        docker pull "$MINER_REGISTRY_NAME:$DOCKER_NAME"
-        docker tag "$MINER_REGISTRY_NAME:$DOCKER_NAME" "$MINER_REGISTRY_NAME:$UPDATE_TAG"
+        docker pull "$MINER_REGISTRY_NAME:$SEMVER_TAG"
+        docker tag "$MINER_REGISTRY_NAME:$SEMVER_TAG" "$MINER_REGISTRY_NAME:$UPDATE_TAG"
         docker push "$MINER_REGISTRY_NAME:$UPDATE_TAG"
 
+        if [["$VERSION_TAG" =~ _GA$]]; then
```
jeffgrunewald, line 95: not sure this is explicitly necessary since this should only ever have a single value in any given build but to prevent any conflicts/mismatch between version shell variables this could also be `if [["$DOCKER_NAME" =~ _GA$]]; then`

jadeallenx (approved):

PR #1441, opened 2022-02-16
Fixing add_gateway call to use add_gateway/4

madninja:
Fixes the dialyzer error on `txns add_gateway`


abhay (approved):

src/jsonrpc/miner_jsonrpc_txn.erl
```diff
@@ -96,3 +99,14 @@ parse_location(#{ <<"lat">> := LatIn,
...
+
+-spec parse_integer(binary() | undefined) -> integer() | undefined.
+parse_integer(undefined) ->
+    undefined;
+parse_integer(Value) when is_binary(Value) ->
+    try 
+        binary_to_integer(Value)
+    catch
+        _:_ ->
+            {error, {invalid_integer, Value}}
```
abhay, line 111: this doesn't return `undefined` like in the spec.

madninja: gah, ok lemme put up another PR.. you may want to leave review comments before approving 😛

PR #1446, opened 2022-02-21
Fix a bug in denylist checking that can cause unintended rollbacks.

ke6jjj:
Denylist freshness relies on comparing the latest available version
with the version the miner currently has. This comparison uses
github tags for versioning, but due to a bug in JSON parsing, any
new version with a _different_ tag than what the miner currently has,
even if it appears to be an _older_ tag, will be accepted as newer,
and will cause the miner to replace the current denylist with the
older version, even if that was unintended.


ke6jjj:
@Vagabond I guess we're going with the other patch, so we can close this. But did the binary() vs. integer() finding make sense?

Vagabond:
> @Vagabond I guess we're going with the other patch, so we can close this. But did the binary() vs. integer() finding make sense?

Yes, you identified the problem, I just fixed it in the "opposite" way.

PR #1457, opened 2022-02-28
allows passing key config values directly to supervisors in tests

jeffgrunewald:
Many common test setups rely on explicitly pre-generating the critical ecc key details, pubkey, ecdh function and signing function, and passing them directly to the application supervisors at startup. This change allows the `miner_keys:keys/0` to accept and passthrough these test-case config overrides directly to the processes that expect them.

### DIFF
--- a/test/miner_ct_utils.erl
+++ b/test/miner_ct_utils.erl
@@ -937,7 +937,7 @@ config_node({Miner, {TCPPort, UDPPort, JSONRPCPort}, ECDH, PubKey, _Addr, SigFun
     ct_rpc:call(Miner, application, set_env, [lager, metadata_whitelist, [poc_id]]),
 
     %% set blockchain configuration
-    Key = {PubKey, ECDH, SigFun},
+    Key = #{pubkey => PubKey, ecdh_fun => ECDH, sig_fun => SigFun},
 
     MinerBaseDir = BaseDir ++ "_" ++ atom_to_list(Miner),
     ct:pal("MinerBaseDir: ~p", [MinerBaseDir]),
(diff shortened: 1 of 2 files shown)


test/miner_ct_utils.erl
```diff
@@ -937,7 +937,7 @@ config_node({Miner, {TCPPort, UDPPort, JSONRPCPort}, ECDH, PubKey, _Addr, SigFun
     ct_rpc:call(Miner, application, set_env, [lager, metadata_whitelist, [poc_id]]),
 
     %% set blockchain configuration
-    Key = {PubKey, ECDH, SigFun},
+    Key = #{pubkey => PubKey, ecdh_fun => ECDH, sig_fun => SigFun},
```
andymck, line 940: sure no side effects of the Key format changing to a map ?

andymck:
Confirming my POC test suite is working again with this change

test/miner_ct_utils.erl
jeffgrunewald: i can't find any; we only ever set that value one time (in the `config_node/2` in the ct_utils setup) and otherwise it's whatever is in sys.config.

this _only_ place i can see this will cause problems is [here](https://github.com/helium/miner/blob/1619185271c85f8d14d2289f05de38c9d20ace8f/src/miner_port_services_sup.erl#L45) which you would _only_ encounter if you're trying to inject pre-built pubkey/signing functions _and_ enabling the embedded gateway-rs and since gateway-rs is supposed to handle all key stuffs when it's turned on i'm not sure that's a valid scenario to worry about anyway

andymck (approved):

PR #1465, opened 2022-03-03
add genesis recreate cli

andymck:
Tested by:
- export current snapshot from test-val1
- load current snapshot onto locally running testnet staked validator
- run recreate CLI command and export new genesis block
- stop local testnet validator and wipe the data directory
- start local testnet validator and load the previously exported gen block
- manually compare datasets against a test validator ( test-val1 )

### DIFF
--- a/src/cli/miner_cli_genesis.erl
+++ b/src/cli/miner_cli_genesis.erl
@@ -22,6 +22,7 @@ register_all_usage() ->
                   [
                    genesis_usage(),
                    genesis_create_usage(),
+                   genesis_recreate_usage(),
                    genesis_forge_usage(),
                    genesis_load_usage(),
                    genesis_export_usage(),
@@ -36,6 +37,7 @@ register_all_cmds() ->
                   [
                    genesis_cmd(),
                    genesis_create_cmd(),
+                   genesis_recreate_cmd(),
                    genesis_forge_cmd(),
                    genesis_load_cmd(),
                    genesis_export_cmd(),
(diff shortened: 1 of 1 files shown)


src/cli/miner_cli_genesis.erl
```diff
@@ -322,6 +325,135 @@ genesis_proof(["genesis", "proof", PrivKeyB58], [], []) ->
...
+    [
+        [["genesis", "recreate", '*', '*'], [], [], fun genesis_recreate/3]
+    ].
+
+genesis_recreate_usage() ->
+    [["genesis", "recreate"],
+        ["genesis recreate <pubkey> <privkey>\n\n",
+            "  recreate a genesis block from the existing ledger data\n"
+            "  includingtxns for all chainvars, gateways, accounts and validators.\n"
+            "  pub and priv key should be base64url encoded binary terms.\n"
```
Vagabond, line 342: Are the keys the 'master key' for the chain?

Vagabond: Would suggest supplying an output directory or file here, vs putting it in whatever the VM's working directory is.

andymck: yeah keys here are the master keys for the chain.  pubkey is then set as 'master_key' in the var txn and the priv key used to create the proof

andymck: NOTE: i simplified the key handling...requirement now is to supply the master key itself as a base58 encoded string rather than its pub and priv components

andymck:
replaced by https://github.com/helium/miner/pull/1467

PR #1467, opened 2022-03-04
add genesis recreate cli

andymck:
Adds CLI and associated helper scripts to export the ledger data to a batch of gen txns, which can then be loaded and used as part of an initial dkg.  Allows a test network such as testnet to be reset whilst maintaining existing data.

Exported data includes accounts, gateways, validators, oracle price & chainvars.

Tested on a local chain via run.sh

### DIFF
--- a/src/cli/miner_cli_genesis.erl
+++ b/src/cli/miner_cli_genesis.erl
@@ -26,7 +26,9 @@ register_all_usage() ->
                    genesis_load_usage(),
                    genesis_export_usage(),
                    genesis_key_usage(),
-                   genesis_proof_usage()
+                   genesis_proof_usage(),
+                   genesis_export_ledger_usage(),
+                   genesis_recreate_ledger_usage()
                   ]).
 
 register_all_cmds() ->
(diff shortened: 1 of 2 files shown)


src/cli/miner_cli_genesis.erl
```diff
@@ -323,6 +326,136 @@ genesis_proof(["genesis", "proof", PrivKeyB58], [], []) ->
...
+                                    [GWAddr | AddrAccc]}
+                        end
+                    end, {[], []}, GWs),
+
+            %% create a consensus group txn
+            {ok, N} = blockchain:config(?num_consensus_members, Ledger),
+            Addrs = [Addr || Addr <- ValAddrs ],
+            SortedAddrs = lists:sort(Addrs),
+            ConsensusAddrs = lists:sublist(SortedAddrs, 1, N),
+            NewConsensusGroupTxn = blockchain_txn_consensus_group_v1:new(ConsensusAddrs, <<>>, 1, 0),
```
Vagabond, line 439: The risk here is that you pick some dead consensus members, how do you boot a chain with this?

andymck: i could check for liveness based on last heartbeat, exclude any which are not recent from the sorted list

andymck: Another option would be to carry across the existing CG members to the new group

Vagabond: I think it'd be better to not include a consensus group txn, and let initial_dkg handle it?

Vagabond: Then you could simply boot the chain on the new validators by giving them the exported genesis txns and the initial consensus group members.

Vagabond: https://github.com/helium/miner/blob/master/src/miner_consensus_mgr.erl#L947-L958

So really you're not making the genesis block directly, you're making the contents of the genesis block

src/cli/miner_cli_genesis.erl
```diff
@@ -40,21 +42,25 @@ register_all_cmds() ->
...
+      "  genesis create <old_genesis_file> <pubkey> <proof> <addrs>         - Create genesis block keeping old ledger transactions.\n"
+      "  genesis forge <pubkey> <key_proof> <addrs>                         - Create genesis block from scratch just with the addresses.\n"
+      "  genesis load <genesis_file>                                        - Load genesis block from file.\n"
+      "  genesis export <path>                                              - Write genesis block to a file.\n"
+      "  genesis key                                                        - create a keypair for use as a master key\n"
+      "  genesis proof <privkey>                                            - create a key proof for adding a master key to the genesis block\n"
+      "  genesis export_ledger <masterkey> <txn_list_output_path>           - export data from ledger as a list of txns including accounts, gateways, validators and chainvars\n"
+      "  genesis recreate_ledger <n> <curve> <path_to_txn_list> <addrs>     - export data from ledger as a list of txns including accounts, gateways, validators and chainvars\n"
```
Vagabond, line 62: This has copied help text from export_ledger

andymck: fixed here: https://github.com/helium/miner/pull/1467/commits/1675824bf67a04674899cb02b6cb681c7afe8601

src/cli/miner_cli_genesis.erl
```diff
@@ -323,6 +329,171 @@ genesis_proof(["genesis", "proof", PrivKeyB58], [], []) ->
...
+                    [clique_status:alert([clique_status:text(io_lib:format("~p", [Reason]))])]
+            end
+end.
+
+%%
+%% genesis recreate_ledger
+%%
+genesis_recreate_ledger_cmd() ->
+    [
+        [["genesis", "recreate_ledger", '*', '*'], [], [], fun genesis_recreate_ledger/3]
```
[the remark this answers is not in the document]
andymck: fixed

PR #1471, opened 2022-03-06
only build embedded gateway in miner releases

jeffgrunewald:
Validator releases have no business building the embedded gateway-rs or semtech-udp project which add significantly to the build time. This change moves the calls to the associated make tasks that build and cleanup the rust projects into the rebar release profiles for miner builds that _do_ need them as embedded dependencies.


jeffgrunewald:
I see that we're still using the testdev profile for the 8 miner builds that get used in the `run.sh` script so assuming that's still an active method of testing the miner code we're not able to remove that profile yet, but i don't see the aws release profile in use anywhere so perhaps we could clean that one up?

evanmcc:
yeah, we don't use the AWS profile anymore, that can be removed.

why do we need this in the test profile?

jeffgrunewald:
> why do we need this in the test profile?

i'm testing that functionality i'd need them to be available for that; specifically in https://github.com/helium/miner/blob/master/test/miner_gateway_mux_integration_SUITE.erl at the moment but could be elsewhere with future tests as well

Vagabond:
it does seem reasonable to leave it in the test profile.

Vagabond (approved):

PR #1476, opened 2022-03-07
Centralize sys.config for container-based testnet miner

ke6jjj:


jeffgrunewald (commented):
why not use/modify the existing `config/docker-testnet.config` instead of creating a new file? the docker_testminer release profile is the only one using the existing config file so there's no need to keep it if you're adding a new file

jeffgrunewald:
Handled in 1491

PR #1479, opened 2022-03-09
Consolidate udp listener port config under radio_device

jeffgrunewald:
Makes it easier for drop-in replacement and update of settings (such as moving from mainnet to testnet or vice versa) by reducing the number of sys.config changes needed to configure the different settings and process trees within the miner.

By consolidating all of the possible variations of a UDP listener to derive their configuration from the existing `miner, radio_device` application env key, the "public interface" for configuring the miner's listener remains the same.

### DIFF
--- a/config/docker.config
+++ b/config/docker.config
@@ -13,8 +13,8 @@
     [
      {gateway_and_mux_enable, false},
      {jsonrpc_ip, {0,0,0,0}}, %% bind jsonrpc to host when in docker container
-     {radio_device, { {0,0,0,0}, 1680, %% change to 1681 when activating mux+gateway-rs
-        {0,0,0,0}, 31341} },
+     {radio_device, { {0,0,0,0}, 1680,
+                      {0,0,0,0}, 31341} },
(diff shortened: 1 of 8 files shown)


ke6jjj (approved):

jeffgrunewald:
If we're trying to make our base config be the "path of least resistance" we should consider removing the radio_device override in `config/docker.config` and just have that release profile inherit the `{{127,0,0,1}, 1680, {127,0,0,1}, 31341}` (or better yet change _that_ profile to deprecate the third and fourth elements of the tuple). _Normally_ it's better practice to have a container listening on `{0,0,0,0}` as we're doing in the config currently, but since we would likely advocate running the container in `--network=host` mode anyway we could encourage this behavior by making the change i'm suggesting. Any maker who _still wants_ to run the miner in a container but _doesn't_ want to run it in host network mode can still always provide a runtime docker overlay of the file but that would be a "discouraged" practice.

Vagabond (approved):

PR #1488, opened 2022-03-16
Change docker to use debian as base image

PaulVMo:
This PR updates the Dockerfile to use Debian and the associated deb package dependencies as the base image. In my testing, the validator docker image performs better when build on Debian as a base image as compared to the existing image based on Alpine. My hypothesis that the optimizations for small size and security within Alpine (and the MUSL C library it uses) result in poorer performance that the glibc based Debian.

For example, I see absorb and commit times that are on the order of 25% faster when running both the Helium Alpine image and on build with this Debian based Dockerfile on the same bare metal 6c/12t Ryzen 5 machine.

Open items:
- [x] Test with hotspot miner build
- [x] Test on ARM
- [x] See if ERTS can be reduced in size - it is significantly larger in the Erlang Debian image as compared to the Alpine one

I have also posted an image built with this Dockerfile for others to use if they wish. It is available here: `ghcr.io/paulvmo/helium-validator:1.8.3` or  `ghcr.io/paulvmo/helium-validator:latest`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,9 @@ RUN apt update \
        wget \
        ${EXTRA_BUILD_APT_PACKAGES} 
 
+# Strip debug symbols from ERTS binarys to reduce size
+RUN scanelf --nobanner -E ET_DYN -BF '%F' --recursive /usr/local/lib/erlang | xargs -r strip --strip-unneeded
+
 # Install Rust toolchain
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-ARG BUILDER_IMAGE=erlang:24-alpine
-ARG RUNNER_IMAGE=alpine
+ARG BUILDER_IMAGE=erlang:24-slim
+ARG RUNNER_IMAGE=debian:bullseye-slim
 FROM ${BUILDER_IMAGE} as deps-compiler
 
 ARG REBAR_DIAGNOSTIC=0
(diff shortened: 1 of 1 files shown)


PaulVMo:
Here is an example of the improvement I'm seeing. This is two validators running on the same Ryzen 5 Pro 4650GE with MVNE storage. The graph captures total absorb and commit time (validate, absorb, and post). I've clipped the chart at 10 seconds to better show the difference as reward blocks can peak at over 60 seconds.
image

Dockerfile
```diff
@@ -55,12 +72,18 @@ RUN wget -O /opt/docker/update/genesis https://snapshots.helium.wtf/genesis.${BU
...
+RUN apt update \
+    && apt install -y \
+       dbus \
+       iproute2 \
+       libncurses6 \
+       libsodium23 \
+       libstdc++6 \
+       ${EXTRA_RUNNER_APT_PACKAGES} \
+    && rm -rf /var/lib/apt/lists \
+    && ln -sf /opt/miner/releases/${VERSION} /config
```
jeffgrunewald, line 86: not sure this symlink statement will succeed until that `/opt/miner` is copied into from the builder copy statement below

PaulVMo: you would think that confirming the target exists is something that `ln` does, however it doesn't seem to. so, for the benefit of reducing a layer in the final image, i moved this up into the other run command. could be easily moved back too if you think that is safer.

PaulVMo:
Some updates:
- Image Size: I was able to remove ~40MB from the image size by stripping debug symbols from the Erlang binaries in the official Erlang base image. They apparently build the official image with the debug compiler flag.
- ARM Validator build - I built and ran this Dockerfile on an Raspberry Pi 4B. while i can't make any performance claims, the validator did run and was able to sync blocks on the RPi.
- Hotspot miner Build - I have been running a cloud (amd64) and a RPi miner with my Dockerfile build for about 24 hours. Both continue to follow the chain and PoC. The performance difference is not as dramatic. The AMD64 one shows ~5% speed up in absorb and post times which I would basically consider within the margin of error. I haven't been able to do a controlled comparison between the ARM hotspot builds.

Vagabond:
Closed in favor of #1780

PR #1490, opened 2022-03-17
move autogen grpc code into proto dep

jeffgrunewald:
This change allows us to consume grpc client modules as application dependencies instead of using the grpcbox plugin to (re)generate them on the fly with each new application build regardless of whether or not the module definitions have changed upstream.


rebar.config
```diff
@@ -40,7 +41,7 @@
               ]}.
 
 {plugins, [
-           {grpcbox_plugin, {git, "https://github.com/andymck/grpcbox_plugin.git", {branch, "andymck/ts-master"}}},
+           {grpcbox_plugin, {git, "https://github.com/andymck/grpcbox_plugin.git", {branch, "andymck/ts-master/combined-opts-and-template-changes"}}},
```
jeffgrunewald, line 44: i'm not clear if this is even necessary anymore or if there's anything else that might be using it.

jeffgrunewald:
_if_ we move in this direction, this should be updated to point to the proto repo's master branch before merging

rebar.config
madninja: I've wondered about this.. did you try removing it?

jwmakesgames (approved):

PR #1491, opened 2022-03-17
Jsc/poc grpc centralize sys config 2

ke6jjj:
Centralize the required sys.config items for the Dockerized testnet miner so that proper operation requires nothing in the overlay sys.config file.


rebar.config
```diff
@@ -303,9 +303,9 @@
                     tools,
                     runtime_tools,
                     recon]},
-                {sys_config, "./config/docker-testnet.config"},
+                {sys_config, "./config/example-overlay.config"},
                 {overlay,
-                  [{copy, "config/sys.config", "config/sys.config"}]},
+                  [{copy, "config/docker-testnetminer.config", "config/sys.config"}]},
```
jeffgrunewald, line 308: i would revert the removal of the `docker-testnet.config` and replace this with `testnet-sys.config` to indicate some parity between `sys.config` and `testnet-sys.config` as the "base config file" that rebar looks to, but i get that it sort of assumes some familiarity with erlang and rebar3 to have that expectation. 

then this can become the primary reference for safe/sane default values for running a miner on testnet and then as with mainnet they need to change those values as needed in the other testnet release profiles (such as test_validators or test_seeds, etc).

PR #1500, opened 2022-03-22
strip mux timestamps to parse logs

jeffgrunewald:
When starting the semtech mux, in order to properly parse logs messages passed back over stdout from the port we need to strip the timestamps generated by the mux to pattern match the log level and dispatch accordingly.


lthiery (approved):

lthiery:
I copy-pasted the logging logic from gateway-rs so this should make the output match, as far as I know.

jeffgrunewald:
Already pulled into another PR

PR #1507, opened 2022-03-25
Bump outbound gossip connections to 4

jadeallenx:


madninja:
@jadeallenx This PR is mostly outdated after the merge of https://github.com/helium/miner/pull/1541, except for the change in outbound connections of seed.config vs sys.config. Can this PR be freshened up?

madninja:
> why so low for seed nodes? 4000 in and 6 out seems like a gossip vacuum rather than seeder
> q2. how many other seed nodes does each seed node connect to in order to keep in sync with changes other seeds receive, and be able to provide the best lookup service? presume answer is `length(maybe_lookup_seed_in_dns(SeedNodes)) - 1`

Actually per https://github.com/helium/blockchain-core/blob/master/src/blockchain_sup.erl#L103-L104 the outbound_gossip_connections should be 18 if max_inbound is 6 right?

madninja:
> > > why so low for seed nodes? 4000 in and 6 out seems like a gossip vacuum rather than seeder
> > > q2. how many other seed nodes does each seed node connect to in order to keep in sync with changes other seeds receive, and be able to provide the best lookup service? presume answer is `length(maybe_lookup_seed_in_dns(SeedNodes)) - 1`
> > 
> > 
> > Actually per https://github.com/helium/blockchain-core/blob/master/src/blockchain_sup.erl#L103-L104 the outbound_gossip_connections should be 18 if max_inbound is 6 right?
> 
> Inbound is 4000, so outbound 6000 by that working lol

why yes it would be..

jadeallenx:
We talked this over and the inbound number is so high because it needs to be high to handle the load for ARP requests which is the seed nodes primary function. The outbound number is really about outbound gossip and it's fine at this level. So we're going to leave it as-is in this PR.

PR #1529, opened 2022-04-04
add RPC and node type to gossip metadata

evanmcc:
For miner and validators, add node type and RPC port url to signed peer metadata.

### DIFF
--- a/src/miner_util.erl
+++ b/src/miner_util.erl
@@ -104,8 +104,18 @@ metadata_fun() ->
         Map = blockchain_worker:signed_metadata_fun(),
         case application:get_env(miner, mode, gateway) of
             validator ->
+                RPC =
+                    case lists:filter(fun libp2p_transport_tcp:is_public/1,
+                                      libp2p_swarm:listen_addrs(blockchain_swarm)) of
+                        [] -> #{};
+                        [First | _] ->
+                            #{<<"rpc_address">> => lists:nth(2, string:tokens(First, "/"))}
+                    end,
+
                 Vsn = element(2, hd(release_handler:which_releases(permanent))),
-                Map#{<<"release_version">> => list_to_binary(Vsn)};
+                Map#{<<"release_version">> => list_to_binary(Vsn),
+                     <<"node_type">> => <<"validator">>},
+                maps:merge(Map, RPC);
             gateway ->
                 FWRelease = case filelib:is_regular(?LSB_FILE) of
                                 true ->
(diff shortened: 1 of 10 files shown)


src/miner_util.erl
```diff
@@ -104,16 +104,30 @@ metadata_fun() ->
         Map = blockchain_worker:signed_metadata_fun(),
         case application:get_env(miner, mode, gateway) of
             validator ->
+                RPC =
+                    case lists:filter(fun libp2p_transport_tcp:is_public/1,
+                                      libp2p_swarm:listen_addrs(blockchain_swarm)) of
+                        [] -> #{};
+                        [First | _] ->
```
jeffgrunewald, line 111: does this filter every realistically return more than one public address? how does a containerized validator typically prevent the container addresses from being tagged as a public listening address and broadcasting it?

PaulVMo:
I think you are looking for GRPC address not JSONRPC address. We don't want to be gossiping around the JSONRPC endpoint which is used for monitoring/control of the node and typical listens on 127.0.0.1. The GRPC is what is used for as the light gateway endpoint, no?
https://github.com/helium/miner/blob/master/config/sys.config#L145-L148

src/miner_util.erl
```diff
@@ -104,16 +104,30 @@ metadata_fun() ->
...
+                            IP = lists:nth(2, string:tokens(First, "/")),
+                            {ok, Port} = application:get_env(miner, jsonrpc_port),
```
jeffgrunewald, line 113:
```suggestion
                            Port = application:get_env(miner, grpc_port, 8080),
```
not sure about the name of the port since it's the server listening port and not a client connection port; i think making it configurable was important though to allow the validator hosters to run on different ports behind the same ip/dns name

src/miner_util.erl
```diff
@@ -104,16 +104,30 @@ metadata_fun() ->
         Map = blockchain_worker:signed_metadata_fun(),
         case application:get_env(miner, mode, gateway) of
             validator ->
+                RPC =
+                    case lists:filter(fun libp2p_transport_tcp:is_public/1,
```
andymck, line 108: maybe better to use libp2p_transport:sort_addrs_with_keys/2 which will put any public ip at the head ?

andymck:
Other than the already commented on grpc port rather than jsonrpc port changes, this should give me what i need in sibyl in order to retrieve the grpc port for the relevant peer.

I will also need to use it to start each validators local grpc listener on the appropriate IP and port

src/miner_util.erl
evanmcc: in the past it has done so enough that I want to handle it explictly, it's a bug, it might have been fixed, but some of this code is gnarly and problematic.

jeffgrunewald (approved):

andymck (approved):

PR #1532, opened 2022-04-07
checking challengees and witnesses against denylist

jeffgrunewald:
within the new miner_poc_mgr, this change allows the validator preparing a challenge to filter out target miners that are present on the denylist and avoid generating challenges to them, as well as filtering out witness reports for them on receipt of witnesses


Vagabond (approved):

andymck (approved):
lg & grpc test passes

PR #1548, opened 2022-04-13
move gateway config concat to before release

jeffgrunewald:
the settings.toml file for the embedded rust gateway needs to be modified for the build network prior to taring up the release so it's copied into the correct directory when the final release is unpacked and then copied over to the runtime container image. in the previous version of this feature, we were modifying the file after the release had been packaged so by the time the settings config was concatenated together that gateway_rs priv directory in `/usr/src/miner` was being ignored by subsequent operations


ke6jjj (approved):
Sounds plausible! All the remains is to test an see if it works.

madninja (approved):

PR #1559, opened 2022-04-15
Disable nat on seeds and validators

jadeallenx:
By default NAT is enabled. For seed nodes and validators this should not be necessary.  The docker validator configuration still leaves NAT enabled though. (Should it be disabled there?)

### DIFF
--- a/config/seed.config
+++ b/config/seed.config
@@ -16,6 +16,7 @@
    {max_inbound_connections, 4000},
    {outbound_gossip_connections, 2},
    {peerbook_update_interval, 60000},
+   {enable_nat, false},
    {gossip_width, 600},
    {relay_limit, 100}
   ]},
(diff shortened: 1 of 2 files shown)


jadeallenx:
Checking on how this affects NAT mapping

kellybyrd:
Does this turn off the public address discovery for a validator? If so, a few comments:
* For Docker validators, the Docker command line in the deployment guide docs use Docker's `bridge` networking which does NAT from the host to the container. I know some validator operators that run host networking, but it's likely not the bulk of them.

* Several popular VPS providers (AWS, GCP, likely others) do a sort of 1:1 NAT between the VM and the outside world. In these cases, I worry turning off NAT will cause those validators to not discover their real public IPs to gossip.

madninja:
I agree with kellybrd here.. I think we should use this as an opportunity to remove NAT and declare a single set of variables for binding listen ipports and public ip with a port for libp2p and a port for grpc. Or something.. it'll need some thought as to all that needs to be captured and explicitly declared by validator operators

PR #1562, opened 2022-04-18
Incorporate debian package enhancements

jadeallenx:
Incorporate #1402 and add the hotfix directory creation


jadeallenx:
Tested on fresh install and upgrade install. Looks good!

PR #1568, opened 2022-04-21
handling non-poc packets in light mode w/o rust gw

jeffgrunewald:
adjustments to the miner_lora_light module to handle routing of packets in either the intended light mode when the light gateway is handling routing of non-poc lorawan packets as well as when the rust gateway is disabled but we're otherwise operating with the light version of the onion and lora servers

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -10,8 +10,7 @@
     port/0,
     position/0,
     location_ok/0,
-    region/0,
-    route/1
+    region/0
 ]).
 
 -export([
@@ -147,19 +146,6 @@ position() ->
 
 -spec location_ok() -> true | false.
 location_ok() ->
-    %% the below code is tested and working.  we're going to roll
-    %% out the metadata update so app users can see their status
-    %% case position() of
-    %%     {error, _Error} ->
-    %%         lager:debug("pos err ~p", [_Error]),
-    %%         false;
-    %%     {ok, _} ->
-    %%         true;
-    %%     %% fix but too far from assert
-    %%     {ok, _, _} ->
-    %%         false
-    %% end.
-
     %% this terrible thing is to fake out dialyzer
     application:get_env(miner, loc_ok_default, true).
 
@@ -629,7 +615,7 @@ handle_packets(_Packets, _Gateway, _RxInstantLocal_us, #state{reg_domain_confirm
     State;
 handle_packets([Packet|Tail], Gateway, RxInstantLocal_us, #state{reg_region = Region, chain = Chain} = State) ->
     Data = base64:decode(maps:get(<<"data">>, Packet)),
-    case ?MODULE:route(Data) of
+    case route(Data) of
         error ->
             ok;
         {onion, Payload} ->
--- a/src/miner_lora_light.erl
+++ b/src/miner_lora_light.erl
@@ -94,8 +94,11 @@
 -define(MAX_TMST_VAL, 4294967295).
 
 -ifdef(TEST).
+-export([route/1]).
+-define(route(Pkt), ?MODULE:route(Pkt)).
 -define(REG_DOMAIN_TIMEOUT, 1000).
 -else.
+-define(route(Pkt), route(Pkt)).
 -define(REG_DOMAIN_TIMEOUT, 30000).
 -endif.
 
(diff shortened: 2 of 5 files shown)


src/miner_lora.erl
```diff
@@ -147,19 +146,6 @@ position() ->
 
 -spec location_ok() -> true | false.
 location_ok() ->
-    %% the below code is tested and working.  we're going to roll
```
jeffgrunewald, line 150: just came across this and looks like it's been commented out for 2 years

src/miner_lora.erl
```diff
@@ -629,7 +615,7 @@ handle_packets(_Packets, _Gateway, _RxInstantLocal_us, #state{reg_domain_confirm
     State;
 handle_packets([Packet|Tail], Gateway, RxInstantLocal_us, #state{reg_region = Region, chain = Chain} = State) ->
     Data = base64:decode(maps:get(<<"data">>, Packet)),
-    case ?MODULE:route(Data) of
+    case route(Data) of
```
jeffgrunewald, line 618: this was only exported for a test previously

test/miner_ct_utils.erl
```diff
@@ -772,6 +772,8 @@ init_per_testcase(Mod, TestCase, Config0) ->
     NumGateways = proplists:get_value(num_gateways, Config, get_config("T", 8)),
     NumConsensusMembers = proplists:get_value(num_consensus_members, Config, get_config("N", 7)),
     LoadChainOnGateways = proplists:get_value(gateways_run_chain, Config, true),
+    GwMuxEnable = proplists:get_value(gateway_and_mux_enable, Config, false),
+    DefaultKeys = proplists:get_value(default_keys, Config, false),
```
jeffgrunewald, line 776: see below where this is consumed; since the miner will auto-generate file-based ecc_compact keys if one is not supplied and i couldn't find any instances where we needed to _specify_ keys for a test (we generate them randomly and inject them directly into the `blockchain, keys` appenv) adding this switch to allow transitioning tests to generate their own keys as part of the miners' startup felt like a non-intrusive way to deprecate that code. this would allow removal of a code path here https://github.com/helium/miner/blob/5aef195d561778d2f8e39a4685045b0f5bd512f7/src/miner_keys.erl#L194 that is only used for testing but does _not_ work when `gateway_and_mux_enable` is true

Vagabond (approved):

src/miner_lora_light.erl
```diff
@@ -490,18 +493,23 @@ handle_packets([Packet|Tail], Gateway, RxInstantLocal_us, #state{reg_region = _R
...
-            %% but in light mode we can just discard non poc packets
-            noop
+        {noop, non_longfi} ->
+            %% operating in light mode and embedded gateway enabled so discard non poc packets
+            lager:debug("Light mode; miner discarding non-Longfi packet ~p", [Packet]),
+            ok;
+        {Type, RoutingInfo} ->
+            %% normally non-Longfi packets in light mode are discarded but this will
+            %% handle when the embedded router is disabled and the miner must route the packet
+            lager:info("Routing ~p", [RoutingInfo]),
```
andymck, line 503: might get noisy

src/miner_lora_light.erl
```diff
@@ -490,18 +493,23 @@ handle_packets([Packet|Tail], Gateway, RxInstantLocal_us, #state{reg_region = _R
...
+            erlang:spawn(fun() -> send_to_router(Type, RoutingInfo, Packet, Region))
```
andymck, line 504: in what scenario would lora light be required to send packets to the router ?

jeffgrunewald: If the gateway-rs is not enabled for whatever reason what is sending packets to the router if not the lora_light module?

jeffgrunewald: i think the original intention was to have the gateway-rs handling all non-poc packets before this change went live, but like with the 4 different versions of h3-dex targeting it's taken longer to fully field test and be comfortable with the rust gateway so it hasn't been "mainstreamed" yet and this would give us the flexibility to fully update either one in either order on mainnet without holding back the other. then we just revert this module to more or less what you originally had once the rust gateway is _always_ part of a standard prod/mainnet hotspot config and build

andymck: k so this covers the possibility where we have light gateways/validator challenges enabled and for whatever reason we want to have miner to route non poc packets.  I dont know why we would want to do that but ok

jeffgrunewald: Personally I don’t think anyone _wants_ to do that, but that’s correct. My hope is we kill it quite quickly

andymck (approved):

PR #1571, opened 2022-04-25
reinstate commented out h3dex vars for POC grpc test

andymck:
These vars were commented out temporarily


andymck:
maybe to alleviate a step here on the validators, the challenger could sign a payload and include in the receipt itself.  Then during absorb each val only has to verify the sig rather than generate the sig

PR #1577, opened 2022-04-27
Add GRPC_IP env to validator config

mbthiery:


abhay:
@mbthiery has this been tested on your fleet of validators?

kellybyrd:
What's the use case where the gRPC IP isn't the same as `LISTEN_ADDRESS`?

mbthiery:
Yes, I've tested it but only when GRPC_IP is the same as LISTEN_ADDRESS. PaulM showed a code snippet where gossiping GRPC_IP would need to change to acknowledge this config.

PR #1578, opened 2022-04-29
use gain field from region params updates

andymck:
Adds gain field to region update streamed responses.  When a GW is asserted or reasserted the region params are sent to the relevant LGW.  Now those updates will include the GWs gain value.  This will then be passed to the onion server light and be used during the tx_power calculation

Related PRs:
https://github.com/helium/proto/pull/131
https://github.com/helium/blockchain-core/pull/1320
https://github.com/helium/sibyl/pull/39


src/poc/miner_onion_server_light.erl
```diff
@@ -325,18 +330,31 @@ try_decrypt(IV, OnionCompactKey, _OnionKeyHash, Tag, CipherText, ECDHFun) ->
...
+        case AssertGain of
+            X when X == undefined orelse ConsiderTxGain == false ->
+                %% No gain on chain or the gain is accounted for in pkt forwarder,
+                %% EIRP = max allowed in region
+                EIRP = trunc(MaxEIRP/10),
+                lager:info("Region: ~p, Gain: ~p, MaxEIRP: ~p, EIRP: ~p",
+                           [Region, undefined, MaxEIRP/10, EIRP]),
+                FinalGain = case AssertGain of
+                           undefined -> 0;
+                           _ -> AssertGain
```
madninja, line 348: Does this need to be `trunc(AsserrGain/10)`?

andymck: I followed same as the non light version ie https://github.com/helium/miner/blob/master/src/poc/miner_onion_server.erl#L502

src/poc/miner_onion_server_light.erl
```diff
@@ -325,18 +330,31 @@ try_decrypt(IV, OnionCompactKey, _OnionKeyHash, Tag, CipherText, ECDHFun) ->
...
-        EIRP = trunc(MaxEIRP/10),
-        lager:info("Region: ~p, Gain: ~p, MaxEIRP: ~p, EIRP: ~p",
-                   [Region, undefined, MaxEIRP/10, EIRP]),
-        {ok, EIRP, EIRP, 0}
+        ConsiderTxGain = application:get_env(miner, consider_tx_gain, true),
+        case AssertGain of
+            X when X == undefined orelse ConsiderTxGain == false ->
+                %% No gain on chain or the gain is accounted for in pkt forwarder,
+                %% EIRP = max allowed in region
+                EIRP = trunc(MaxEIRP/10),
```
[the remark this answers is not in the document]
andymck: @vihu in case it's not obvious the same /10 has always been happening on the regular miner kora module. Are you content with removing it there also then..even tho it's what is in use atm on hotspots?

vihu: Yeah, we retraced the steps and it seems to be mirroring the previous version. So I think we're likely okay.

vihu (approved):

PR #1579, opened 2022-04-29
setting sc downlink handler per active lora mod

jeffgrunewald:
in the transitional phase where we are running under validator challenges/hotspots are no longer following the chain but we have not enabled the embedded gateway-rs packet router we will be using the `miner_lora_light` module to route lora packets received by the gateway. this change allows the correct / active lora module handling packets to register itself with the blockchain_state_channel_client handler so downlink packets can be returned to the source device correctly regardless of which module is currently active.

### DIFF
--- a/test/miner_ct_utils.erl
+++ b/test/miner_ct_utils.erl
@@ -1164,7 +1164,6 @@ config_node({Miner, {GwApiPort, UDPPort, JSONRPCPort}, ECDH, PubKey, _Addr, SigF
     ct_rpc:call(Miner, application, set_env, [blockchain, max_inbound_connections, TotalMiners*2]),
     ct_rpc:call(Miner, application, set_env, [blockchain, outbound_gossip_connections, TotalMiners]),
     ct_rpc:call(Miner, application, set_env, [blockchain, sync_cooldown_time, 5]),
-    %% ct_rpc:call(Miner, application, set_env, [blockchain, sc_client_handler, miner_test_sc_client_handler]),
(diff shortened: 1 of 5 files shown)


test/miner_ct_utils.erl
```diff
@@ -1164,7 +1164,6 @@ config_node({Miner, {GwApiPort, UDPPort, JSONRPCPort}, ECDH, PubKey, _Addr, SigF
     ct_rpc:call(Miner, application, set_env, [blockchain, max_inbound_connections, TotalMiners*2]),
     ct_rpc:call(Miner, application, set_env, [blockchain, outbound_gossip_connections, TotalMiners]),
     ct_rpc:call(Miner, application, set_env, [blockchain, sync_cooldown_time, 5]),
-    %% ct_rpc:call(Miner, application, set_env, [blockchain, sc_client_handler, miner_test_sc_client_handler]),
```
jeffgrunewald, line 1167: git indicates this line has been commented out since at least 2/2021 so this felt like the easiest solution to avoid breaking any tests using this setting since the lora modules would just immediately stomp on this value when the miner application is started below. if we'd like to keep this in i can also wrap a case around the checks in the lora modules (in a helper function) to only set the value to the active lora module _if_ it's not already defined. it feels janky but since both those modules try to duplicate each others' efforts i don't _think_ it could result in adverse race conditions and it would only really take effect if someone proactively set the value like this or in sys.config

andymck (approved):
good find. lgtm

PR #1586, opened 2022-05-05
update testnet config to reduce crosstalk with mainnet

joecaswell:
- do not use dns for seeds (that only provides mainnet seeds)
- add both testnet seeds to seed_nodes
- add aliases for the testnet seeds
- remove aliases for mainnet nodes

### DIFF
--- a/config/testnet-sys.config
+++ b/config/testnet-sys.config
@@ -25,7 +25,7 @@
   ]},
  {libp2p,
   [
-   {use_dns_for_seeds, true},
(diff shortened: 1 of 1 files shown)


ke6jjj (approved):

abhay:
Nice. @joecaswell. Can we also have equivalent ones in 

* https://github.com/helium/blockchain-node
* https://github.com/helium/blockchain-etl
* https://github.com/helium/router

Maybe missing some

PR #1590, opened 2022-05-06
Andymck/fix grpc leak

andymck:
Related PRs:

NOTE: DONT MERGE THIS UNTIL THE TWO PRS BELOW ARE MERGED AND DEPS UPDATED HERE!

https://github.com/helium/blockchain-core/pull/1332
https://github.com/helium/sibyl/pull/41

Addresses three issues

1.  (clientside fix )When a previously established grpc stream goes down, the retry happens immediately.  This change adds a simple 5 sec delay between retries  

2. Adds a connection stability check.  Once connected to a validator the number of times any stream disconnects will be monitored.  If the number of disconnects within the specified period exceeds a max, then we will consider the validator to be unstable and switch to a new validator

3. ( serverside fix ) Pulls in an updated chatterbox and grpcbox with fixes to address a process leak ( explained below )

GrpcBox Process Leak Outline
-----------------------------
when grpcbox ends a stream ( either normally or upon hitting an exception ) it correctly sends the end_stream flag and the trailers.  This would put the server side endpoint into a half_open_local state ( the side which sends the end_stream frame transitions to half_open_local with the other endpoint transitioning to half_open_remote ).

As per the http2 spec, when an endpoint ( client or server ) transitions to the half open state, either side can then subsequently send the rst_stream frame in order to close the stream.  In the case of our memory leak no endpoint was doing that and so the stream process on the server side would remain alive.  The client was fine as it handled the end_stream flag received from the server and terminated.

Given we dont want to be at the mercy of a client sending the rst_stream frame I have modified grpcbox so that any time it initiates an end_stream flag and enters the half_open_local state it will also send a rst_stream frame.  This ensures the server side always cleans up.  This feels like best practice from the application side.

We dont have an issue when the server side enters a half_open_remote state ( client sends the end flag) as when that occurs grpcbox will subequently send trailers which will result in the stream moving to a closed state and the process terminating.

chatterbox fix: 

https://github.com/andymck/chatterbox/commit/f42b289f7035752f3ef535a81bcc6fa418ea24b0
https://github.com/andymck/chatterbox/commit/9055e611d509eec2f182bbf52a54ec821b17ac59

grpcbox fix:  https://github.com/andymck/grpcbox/commit/37268026aeb18f907197ebb2cabc59b6c000e9c3


jeffgrunewald:
if we're accepting that we're not immediately able to reconnect after disconnects does it make sense to add a `disconnected` state to the statem and drop into that for these types of scenarios?

src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -381,6 +381,11 @@ connected({call, From}, connection, #data{connection = Connection} = Data) ->
 connected(info, {'DOWN', _Ref, process, _, _Reason} = Event, Data) ->
     lager:warning("got down event ~p", [Event]),
     %% handle down msgs, such as from our streams or validator connection
+    %% dont handle it right away, instead wait some time first
+    erlang:send_after(5000, self(), {delayed_down_event, Event}),
+    {keep_state, Data};
+%%    handle_down_event(connected, Event, Data);
```
evanmcc, line 387: delete

evanmcc (commented):
would like to see a retry limit like you mentioned in the description.

evanmcc:
this is looking good but a lot of these logs should be reverted to debug

andymck:
> if we're accepting that we're not immediately able to reconnect after disconnects does it make sense to add a `disconnected` state to the statem and drop into that for these types of scenarios?

the setup state is kinda like a disconnected state...the issue with a purely disconnected state is what does it represent.  We have a single validator connection and then over that connection we have at least 3 streams any of which could be connected or disconnected at any one time.  It becomes difficult to represent all those via a disconnected state.

evanmcc (approved):

PR #1593, opened 2022-05-10
Introduce metrics framework based on telemetry events

jeffgrunewald:
Consumes the following upstream PRs to add telemetry events to the underlying blockchain-core and grpcbox libraries that will emit telemetry events related to block production and grpc connections/request latencies.

The metrics framework is designed to be extensible and consume any telemetry events being generated within the miner software or any upstream dependencies and by added to the expected prometheus scrape target endpoint by updating the handler function with a new function head (`src/metrics/miner_metrics_server.erl`), and by updating the event specs in the metrics header file (`include/metrics.hrl`) and then adding the desired groups of metrics to your sysconfig profile.

Depends on merges of:
https://github.com/andymck/grpcbox/pull/3
https://github.com/helium/blockchain-core/pull/1336
https://github.com/helium/sibyl/pull/42

and then subsequent cleanup of the rebar.config/rebar.lock files.


src/miner_critical_sup.erl
```diff
@@ -105,7 +105,10 @@ init(_Opts) ->
         end,
 
 
-    ChildSpecs0 = [?SUP(blockchain_sup, [BlockchainOpts])] ++ ConsensusMgr,
+    ChildSpecs0 = [
```
jadeallenx, line 108: Probably belongs in the `restart_sup` don't you think?

evanmcc: it doesn't actually; failures in the critical_sup will stop the whole system, so this definitely belongs in the restart_sup.

jeffgrunewald: I meant the metrics functionality works the same either way

PR #1594, opened 2022-05-10
Use the correct chain var for bloom filter reset

mawdegroot:
When running with validator challenges the bloom filter is still reset using `poc_challenge_interval` which is not used with validator challenges. Change to the correct chain variable and use `GatewayCount / PocChallengeRate` to mimic old bloom filter behavior.

`blockchain_ledger_v1:gateway_count` is quite slow so adding it to the record to cache the result for the duration of one interval length.

### DIFF
--- a/src/poc/miner_poc_mgr.erl
+++ b/src/poc/miner_poc_mgr.erl
@@ -51,6 +51,7 @@
 -record(addr_hash_filter, {
     start :: pos_integer(),
     height :: pos_integer(),
+    interval :: pos_integer(),
     byte_size :: pos_integer(),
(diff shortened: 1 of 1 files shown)


abhay:
This likely needs some testing but it also needs a guard around rollback scenario if we (unlikely) roll back the light hotspot activation.

mawdegroot:
From what I understood the `miner_poc_mgr` is only used with validator challenges, if we roll back to hotspot challenges we will fall back to `miner_poc_statem` so a guard shouldn't be necessary.

How do I best test this?

PR #1608, opened 2022-05-12
Add a delay between send_report retries

joecaswell:
Using a configurable delay with a default of 30s, this should give the validator time to catch up.

### DIFF
--- a/src/poc/miner_poc_grpc_client_statem.erl
+++ b/src/poc/miner_poc_grpc_client_statem.erl
@@ -61,6 +61,8 @@
 
 %% delay between validator reconnects attempts
 -define(VALIDATOR_RECONNECT_DELAY, 5000).
+-define(DEFAULT_GRPC_SEND_REPORT_RETRY_DELAY, 30000).
+-define(DEFAULT_GRPC_SEND_REPORT_RETRY_ATTEMPTS, 20).
 -ifdef(TEST).
 %% delay between active check retries
 %% these checks determine where to proceed with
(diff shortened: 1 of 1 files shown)


src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -611,10 +622,10 @@ do_send_report(Req, ReportType, Report, OnionKeyHash, Connection, RetryAttempts)
                 {ok, _} ->
                     ok;
                 _ ->
-                    ?MODULE:send_report(ReportType, Report, OnionKeyHash, RetryAttempts - 1)
+                    ?MODULE:retry_report(ReportType, Report, OnionKeyHash, RetryAttempts - 1)
```
jeffgrunewald, line 625: Why not just put a `timer:sleep/1` in here and then call the existing `send_report/4` function?

jeffgrunewald: If you’re looking to get the gen_statem out of this function and not sit in the timer then I’d have this guy schedule the `send_after` and interpose a small handle_info that just catches that message and then calls `send_report/4` passing down the args and decrementing the retry. Mimicking a cast in a simple send message with the magic `'$gen_cast'` atom seems brittle to me

joecaswell: that's reasonable

jeffgrunewald (approved):

src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -611,13 +614,17 @@ do_send_report(Req, ReportType, Report, OnionKeyHash, Connection, RetryAttempts)
                 {ok, _} ->
                     ok;
                 _ ->
-                    ?MODULE:send_report(ReportType, Report, OnionKeyHash, RetryAttempts - 1)
+                    retry_report({send_report, ReportType, Report, OnionKeyHash, RetryAttempts - 1})
```
abhay, line 617: could we add some failure logging here too? useful for debug and we can turn it down later.

src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -611,13 +614,17 @@ do_send_report(Req, ReportType, Report, OnionKeyHash, Connection, RetryAttempts)
...
             end;
         {error, _Reason} ->
-            ?MODULE:send_report(ReportType, Report, OnionKeyHash, RetryAttempts - 1)
+            retry_report({send_report, ReportType, Report, OnionKeyHash, RetryAttempts - 1})
```
abhay, line 620: also failure logging here.

joecaswell: That is simple enough

joecaswell:
squashing commits

vihu (approved):

PR #1625, opened 2022-05-13
grpc poc statem block age timeout

jeffgrunewald:
After running the grpc poc infrastructure at scale for a few days we've found hotspots can get into the undesirable state of being connected to a validator that is maintaining stable grpc streams but is not keeping up with the chain or may not have the latest blocks available to validate the connected hotspots' challenges.

This change ensures that hotspots are able to bail off a poorly performing validator and search for a new one if the connected validator is not keeping up with the chain to roughly within 15 min (assuming 1 block/minute for a total block age allowance of 15 blocks behind).

### DIFF
--- a/src/poc/miner_poc_grpc_client_statem.erl
+++ b/src/poc/miner_poc_grpc_client_statem.erl
@@ -52,7 +52,8 @@
     val_grpc_port,
     check_target_req_timer,
     down_events_in_period = 0,
-    stability_check_timer
+    stability_check_timer,
+    block_age_timer
 }).
 
 %% these are config vars the miner is interested in, if they change we
@@ -83,6 +84,9 @@
 %% max number of stream down events  within the STREAM_STABILITY_CHECK_TIMEOUT period
 %% after which we will select a new validator
 -define(MAX_DOWN_EVENTS_IN_PERIOD, (?STREAM_STABILITY_CHECK_TIMEOUT / ?RECONNECT_DELAY) - 2).
+-define(BLOCK_AGE_TIMEOUT, 300000).
+%% Maximum block age returned by a connected validator before triggering an instability reconnect
+-define(MAX_BLOCK_AGE, 900000).
 %% ets table name for check target reqs cache
 -define(CHECK_TARGET_REQS, check_target_reqs).
 -type data() :: #data{}.
@@ -241,13 +245,16 @@ setup(
             %% once we have a validator connection established
             %% fire a timer to check how stable the streams are behaving
             catch erlang:cancel_timer(Data#data.stability_check_timer),
-            TRef = erlang:send_after(?STREAM_STABILITY_CHECK_TIMEOUT, self(), stability_check),
+            catch erlang:cancel_timer(Data#data.block_age_timer),
+            SCTRef = erlang:send_after(?STREAM_STABILITY_CHECK_TIMEOUT, self(), stability_check),
+            BACTRef = erlang:send_after(?BLOCK_AGE_TIMEOUT, self(), block_age_check),
             {keep_state,
                 Data#data{
                     connection = Connection,
                     connection_pid = ConnectionPid,
                     conn_monitor_ref = M,
-                    stability_check_timer = TRef,
+                    stability_check_timer = SCTRef,
+                    block_age_timer = BACTRef,
                     down_events_in_period = 0
                 },
                 [{next_event, info, fetch_config}]};
@@ -345,6 +352,8 @@ setup(info, {delayed_down_event, Event}, Data) ->
     handle_down_event(connected, Event, Data);
 setup(info, stability_check, Data) ->
     handle_stability_check(setup, Data);
+setup(info, block_age_check, Data) ->
+    handle_block_age_check(setup, Data);
 
 setup({call, From}, _Msg, Data) ->
     %% return an error for any call msgs whilst in setup state
@@ -420,7 +429,9 @@ connected(info, process_queued_check_target_reqs, Data) ->
     Ref = erlang:send_after(?CHECK_TARGET_REQ_DELAY, self(), process_queued_check_target_reqs),
     {keep_state, Data#data{check_target_req_timer = Ref}};
 connected(info, stability_check, Data) ->
-    handle_stability_check(info, Data);
+    handle_stability_check(connected, Data);
+connected(info, block_age_check, Data) ->
+    handle_block_age_check(connected, Data);
 connected(_EventType, _Msg, Data) ->
     lager:debug(
         "unhandled event whilst in ~p state: Type: ~p, Msg: ~p",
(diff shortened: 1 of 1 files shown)


src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -1151,6 +1162,57 @@ handle_stability_check(CurState, Data = #data{down_events_in_period = NumDownEve
...
+        lager:info("Send block age unary request", []),
+        case grpc_client_custom:unary(
+            Connection,
+            build_config_req([]),
+            'helium.gateway',
+            config,
+            gateway_miner_client_pb,
+            [{callback_mod, miner_poc_grpc_client_handler}]
+        ) of
+            {ok, #{http_status := 200, result := #gateway_resp_v1_pb{block_age = BlockAge}}} where is_integer(BlockAge) ->
```
vihu, line 1204:
```suggestion
            {ok, #{http_status := 200, result := #gateway_resp_v1_pb{block_age = BlockAge}}} when is_integer(BlockAge) ->
```

src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -1151,6 +1162,57 @@ handle_stability_check(CurState, Data = #data{down_events_in_period = NumDownEve
...
+            end;
+        _ ->
+            %% connected validator seems to be keeping up; once more around the loop
+            TRef = erlang:send_after(?BLOCK_AGE_TIMEOUT, self(), block_age_check),
+            {keep_state, Data#data{block_age_timer = TRef}}
+    end.
+
+send_block_age_req(Connection) ->
+    try
+        lager:info("Send block age unary request", []),
```
jadeallenx, line 1195: Is this going to be really spammy? Maybe it should be `debug`

jeffgrunewald: quite possibly; we have a lot of those at the moment where we've dialed up the logging. although if this is unlikely to make a difference having the additional logging might not be valuable
for now i was just mirroring the log level of the statements in the `send_grpc_unary_req` function

jeffgrunewald: looking again, i don't see a ton of value in this message being at the `info` level other than as a "i'm still alive doing what i'm doing" so i turned it down to debug

src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -1151,6 +1163,60 @@ handle_stability_check(CurState, Data = #data{down_events_in_period = NumDownEve
...
+                {error, grpc_request_failed}
+        end
+    catch
+        _Class:_Error:_Stack ->
+            lager:warning("send block age unary request failed: ~p, ~p, ~p", [_Class, _Error, _Stack]),
+            {error, req_failed}
+    end.
+
+block_age_timeout() ->
+    ?BLOCK_AGE_TIMEOUT + rand:uniform(?BLOCK_AGE_TIMEOUT).
```
vihu, line 1218: Should there be a `timer:seconds()` around this?

vihu: oh nvm, read the wrong constant above

vihu (approved):

abhay:
I think this change can cause a thundering herd on the seed validators in the case that we have a chain halt. If i'm reading it correctly, after 15 (+random) minutes of a chain halt, each hotspot will loop and try to ask for a new validator to connect to. wonder if we should keep a reconnect count and a backoff?

cc: @jeffgrunewald @vihu @evanmcc

src/poc/miner_poc_grpc_client_statem.erl
```diff
@@ -1151,6 +1163,60 @@ handle_stability_check(CurState, Data = #data{down_events_in_period = NumDownEve
...
+            Data1 = Data#data{down_events_in_period = 0},
+            lager:warning("validator block age ~p exceeded maximum allowed; switching validators", [BlockAge]),
+            case CurState of
+                setup -> {repeat_state, Data1};
+                _ -> {next_state, setup, Data1}
+            end;
+        {error, _Reason} ->
+            %% validator, you had one job...
+            Data1 = Data#data{down_events_in_period = 0},
+            lager:warning("validator block age request failed; switching validators", [_Reason]),
```
BenoitDuffez, line 1183: `~p`?

jeffgrunewald: oh dang, good catch. that'll get converted to a "FORMAT ERROR" message so it won't be as pretty as intended but it'll be harmless otherwise until the next release

jeffgrunewald:
> I think this change can cause a thundering herd on the seed validators in the case that we have a chain halt. If i'm reading it correctly, after 15 (+random) minutes of a chain halt, each hotspot will loop and try to ask for a new validator to connect to. wonder if we should keep a reconnect count and a backoff?
> 
> cc: @jeffgrunewald @vihu @evanmcc

I think @Vagabond made a comment earlier about hotspots having seen a bunch of validators over the course of operation; do we want to have them always go back to the seed validators in the event of reconnect or instead maybe have them cache the last N validators they've talked to (maybe the last N connections and challengers or something) and start walking the chain back along that line, trying one at a time until they get a connection and only after they've exhausted _that_ list go hit up the seed validators for more?

for that matter, we could warm said cache by having the request to the seed validators return 3-5 random validators instead of just 1 and those would populate the cache on first boot of the miner?

abhay:
> > I think this change can cause a thundering herd on the seed validators in the case that we have a chain halt. If i'm reading it correctly, after 15 (+random) minutes of a chain halt, each hotspot will loop and try to ask for a new validator to connect to. wonder if we should keep a reconnect count and a backoff?
> > cc: @jeffgrunewald @vihu @evanmcc
> 
> I think @Vagabond made a comment earlier about hotspots having seen a bunch of validators over the course of operation; do we want to have them always go back to the seed validators in the event of reconnect or instead maybe have them cache the last N validators they've talked to (maybe the last N connections and challengers or something) and start walking the chain back along that line, trying one at a time until they get a connection and only after they've exhausted _that_ list go hit up the seed validators for more?
> 
> for that matter, we could warm said cache by having the request to the seed validators return 3-5 random validators instead of just 1 and those would populate the cache on first boot of the miner?

all seems reasonable.

madninja:
> > > I think this change can cause a thundering herd on the seed validators in the case that we have a chain halt. If i'm reading it correctly, after 15 (+random) minutes of a chain halt, each hotspot will loop and try to ask for a new validator to connect to. wonder if we should keep a reconnect count and a backoff?
> > > cc: @jeffgrunewald @vihu @evanmcc
> > 
> > 
> > I think @Vagabond made a comment earlier about hotspots having seen a bunch of validators over the course of operation; do we want to have them always go back to the seed validators in the event of reconnect or instead maybe have them cache the last N validators they've talked to (maybe the last N connections and challengers or something) and start walking the chain back along that line, trying one at a time until they get a connection and only after they've exhausted _that_ list go hit up the seed validators for more?
> > for that matter, we could warm said cache by having the request to the seed validators return 3-5 random validators instead of just 1 and those would populate the cache on first boot of the miner?
> 
> all seems reasonable.

I think it's a bit of premature optimization, personally

PR #1633, opened 2022-05-16
ara/reimplement-shuffle-of-witnesses

anthonyra:
Upon some code review, I noticed that with challenges by validators we no longer shuffled then truncated the witnesses collected during PoC. This reimplements the old PoC logic in regards to witnesses before HIP-55.


andymck:
@anthonyra thank you for this.  The branch fails to compile:

===> Compiling src/poc/miner_poc_mgr.erl failed
src/poc/miner_poc_mgr.erl:715:85: variable 'Ledger' is unbound

andymck (approved):
Looks good, POC CT is passing.

PR #1642, opened 2022-05-16
audit 134

jeffgrunewald:
Adds the audit record for chainvar 134 upping the PoC rate interval from 400 to 700.

Includes a small cleanup to a broken logging format string and a correction to the header line of the audit for var 133


madninja (changes requested):
don't put non audit stuff in the PR please?

madninja (approved):

PR #1654, opened 2022-05-17
various check target request improvements

andymck:
- retry check target requests if grpc req fails
- serialise check target requests, dont spawn off
- remove sync call during check target to blockchain_swarm:get_keys/0
- prevent connection close calls from crashing, if already dead


jeffgrunewald (approved):
Approved, barring any changes to the pubkey/bin -> animal name logging you may want to remove

PR #1655, opened 2022-05-17
Attempt count/log validation results by type per round in sidecar

jeffgrunewald:
Institute rudimentary validation counts on txns per hbbft round in the validation sidecar. Each round, internally collect a count of txns by type, grouped by validation result and emit the counts at the reset of each round in a log message

### DIFF
--- a/src/miner_hbbft_sidecar.erl
+++ b/src/miner_hbbft_sidecar.erl
@@ -33,10 +33,11 @@
 %% txns that do not appear naturally
 -define(InvalidTxns, [blockchain_txn_reward_v1, blockchain_txn_reward_v2]).
 
+-define(SUBMITTED_TXNS, submitted_txns).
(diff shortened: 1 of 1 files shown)


jeffgrunewald:
Please note the placement of the "Submitted Txns" submission metric within the `submit` handle_call/3 but only at the point where the txn is about to begin validation. This means that txns submitted before we have a chain or a group or txns that are rejected outright by various basic invalidations of their type are _not_ counted. This may or may not be the desired metric but my thinking was, if the intention is to compare txns submitted for validation vs. the results of that validation process than this is the correct point to count submissions for the most "apples to apples" comparison; avoid skewing results by also counting failed submissions due to a chain not being set or misconfigured node submitting txns it shouldn't.

evanmcc (approved):

PR #1664, opened 2022-05-18
add as923_1b region

mikev:
AS923_1B defines a Region supporting Malaysia. It is illegal in Malaysia to use any frequency above 924 Mhz, hence the reason for this new Region. The AS923_1B defines Uplink frequencies at [922.0, 922.2, 922.4, 922.6, 922.8, 923.0, 923.2, 923.4]

This change depends on blockchain-core changes - https://github.com/helium/blockchain-core/pull/1355


lthiery:
This looks right to me but I am from an expert in the miner's poc codebase. I double checked the frequencies and bandwidths and they look right, though.

What's the meaning of the first value in the Spreading Factor data?
```
         [{25, 'SF12'}, {25, 'SF11'}, {25, 'SF10'}, {67, 'SF9'}, {139, 'SF8'}, {256, 'SF7'}]}
```

ke6jjj:
@lthiery I think the first value in the tuple is the maximum allowable transmit packet size, which is derived from the maximum time on air multiplied by the bandwidth of that SF.

lthiery:
> @lthiery I think the first value in the tuple is the maximum allowable transmit packet size, which is derived from the maximum time on air multiplied by the bandwidth of that SF.

Got it. That makes sense

Vagabond (approved):
This looks harmless

oberchoo:
Looking forward it to being merged soon.

oberchoo:
Great! Thanks everyone

PR #1674, opened 2022-05-19
ara/update-queue-time-once

anthonyra:
I noticed that the queue time was "updated" every time that a hotspot got the "queue_poc" or "req_failed" response from `check_target` with this queue time being used in `process_check_target_reqs` it should be only update on first entrance into the queue. This will ensure the past due removal works as intended.

So prior to re-inserting the req into the queue, see if there's one already. If there is, use the already given queue time for the update. Not sure, if anything outside of it should change... maybe there's a better approach here also just what came to mind for me at the time.


anthonyra:
Scratch this, it's controlled via a flag

anthonyra:
Scratch this, it's controlled via a flag

PR #1678, opened 2022-05-20
harmonize local keys GC cadence with ledger key proposals GC interval

andymck:
When generating ephemeral keys for a POC, a validator will generate key pairs.   The public key will be proposed in the its heartbeats and end up on the ledger proposed pocs CF, whilst the private key will be saved to a local rocks DB via miner_poc_mgr.

These two keys are intrinsically linked, if a public key is selected/promoted then in order for the POC to initiate the owning validator must still have the associated private key.  However the GC implementation for these key stores were running at a different cadence...the public keys would have a GC run every 101 blocks whilst the private keys would have a GC run every 50 blocks.

This allowed for the possibility of a local private key being removed prior to the public key. Should the public key then be selected the POC would fail.

This fix harmonises the cadence of both.

TODO: is a chain var necessary here ?  I dont believe so as it impacts only local data....but that local data does influence if a POC initiates or not ????

### DIFF
--- a/src/poc/miner_poc_mgr.erl
+++ b/src/poc/miner_poc_mgr.erl
@@ -336,8 +336,10 @@ handle_add_block_event(POCChallengeType, BlockHash, Ledger, State) when POCChall
             %% take care of GC
             ok = purge_local_pocs(Block, Ledger, State),
 
-            %% GC local pocs keys every 50 blocks
-            case BlockHeight rem 50 == 0 of
+            %% GC local pocs keys every 101 blocks
+            %% this is tied to the GC cadence window
+            %% for the key propopsals in ledger v1
+            case BlockHeight rem 101 == 0 of
                 true ->
                     ok = purge_local_poc_keys(BlockHeight, Ledger, State);
                 false ->


src/poc/miner_poc_mgr.erl
```diff
@@ -336,8 +336,10 @@ handle_add_block_event(POCChallengeType, BlockHash, Ledger, State) when POCChall
             %% take care of GC
             ok = purge_local_pocs(Block, Ledger, State),
 
-            %% GC local pocs keys every 50 blocks
-            case BlockHeight rem 50 == 0 of
+            %% GC local pocs keys every 101 blocks
+            %% this is tied to the GC cadence window
+            %% for the key propopsals in ledger v1
+            case BlockHeight rem 101 == 0 of
```
madninja, line 342: is there a way to tie it to a constant in ledger v1 then?

madninja (commented):
Just one comment which may be too much right now, and I do see some stern warnings in ledger_v1 to not change that value. 

If not shareable in an existing header, this looks fine

src/poc/miner_poc_mgr.erl
evanmcc: we could make this an exported ledger function returning a static number if we wanted.

evanmcc (approved):
This should work, won't require a chain var.  The only things that require chain vars in the miner are things involved in block production directly (which I suppose we could push back into core for consistency), rather than through transactions.

src/poc/miner_poc_mgr.erl
jeffgrunewald: maybe rip this block out into an exported function on the ledger and call it in both places? https://github.com/helium/blockchain-core/blob/d78c0abe8beaea8a487ccbb9ebe5646f8365a88f/src/ledger/v1/blockchain_ledger_v1.erl#L2366

madninja: > maybe rip this block out into an exported function on the ledger and call it in both places? https://github.com/helium/blockchain-core/blob/d78c0abe8beaea8a487ccbb9ebe5646f8365a88f/src/ledger/v1/blockchain_ledger_v1.erl#L2366

worth putting on the list of cleanup tasks. I'll get this merged in the meantime

PR #1693, opened 2022-05-24
consume poc gc interval exported from ledger

jeffgrunewald:
consume ledger function in core for determining poc gc interval in blocks instead of relying on the value in both repos being in sync


src/poc/miner_poc_mgr.erl
```diff
@@ -336,10 +336,9 @@ handle_add_block_event(POCChallengeType, BlockHash, Ledger, State) when POCChall
             %% take care of GC
             ok = purge_local_pocs(Block, Ledger, State),
 
-            %% GC local pocs keys every 101 blocks
-            %% this is tied to the GC cadence window
-            %% for the key propopsals in ledger v1
-            case BlockHeight rem 101 == 0 of
+            %% GC local pocs keys based on the GC cadence window in blocks
+            %% for the key propopsals in ledger v1; defaults to 101
+            case BlockHeight rem blockchain_ledger_v1:poc_gc_interval(Ledger) == 0 of
```
andymck, line 341: this would need tied to the `poc_proposal_gc_window_check` chain var, should follow same logic as here:

https://github.com/helium/blockchain-core/blob/master/src/ledger/v1/blockchain_ledger_v1.erl#L2365

jeffgrunewald: It is; this is the PR that introduces the function used here in miner: https://github.com/helium/blockchain-core/pull/1364/files

andymck: ah k my bad, i misread the ledger call.

andymck (approved):
lg.  can you link to the dependant core PR in the description, helps avoid that being missed

madninja (approved):

PR #1703, opened 2022-05-28
accommodate devnet builds in the build pipeline

jeffgrunewald:
create and push images to a devnet repository based on git tags similar to testnet


jadeallenx (approved):

JayKickliter (approved):
Thanks for commenting all the different build paths in the yml

PR #1705, opened 2022-05-30
stream related grpc client fixes

andymck:
1.  Adds a monitor to the stream process of the parent connection process. If that terminates then shutdown the stream process
2.  Specifies a timeout for unary grpc requests.  If no timeout is specified the grpc client lib would default to infinity
3.  Adds a query to find txns by onion key hash
4.  Logging tweaks

### DIFF
--- a/src/miner_query.erl
+++ b/src/miner_query.erl
@@ -13,7 +13,9 @@
          txns/2, txns/3,
          blocks/2,
          lookup_txns_by_hash/2,
-         lookup_txns_by_type/2]).
+         lookup_txns_by_type/2,
+         lookup_txns_by_onionhash/2
+]).
 
 poc_analyze(_Start, _End) ->
     ok.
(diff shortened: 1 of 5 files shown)


src/miner_query.erl
```diff
@@ -90,3 +92,25 @@ lookup_txns_by_hash(LastXBlocks, TxnHash) ->
         end,
         Current - LastXBlocks, Current, C,
         [])).
+
+lookup_txns_by_onionhash(LastXBlocks, OnionHash) ->
```
madninja, line 96: this isn not used anywhere?

src/poc/miner_poc_mgr.erl
```diff
@@ -544,8 +546,8 @@ initialize_poc(BlockHash, POCStartHeight, Keys, Vars, Ledger, #state{pub_key = C
                         start_height = POCStartHeight
                     },
                     ok = write_local_poc(LocalPOC, State),
-                    lager:info("started poc for challengeraddr ~p, target: ~p, onionhash ~p",
-                        [?TO_ANIMAL_NAME(Challenger), OnionKeyHash]),
+                    lager:info("started poc at blockheight ~p for challengeraddr ~p, onionhash ~p, target: ~p",
```
madninja, line 549: fyi, I've started calling it poc_id instead of onion_key_hash

andymck: on the miner side, poc_id is the unhashed version of the onion key

madninja: then unhashed version is the poc_public_key :-)

madninja (approved):

PR #1712, opened 2022-06-02
disable the mux to bypass an issue that requires testing

jeffgrunewald:
hotfix to bypass a packet routing issue in the embedded rust gateway while a pending fix is validated


config/docker.config
```diff
@@ -15,6 +15,5 @@
      {radio_device, { {0,0,0,0}, 1680,
                       {0,0,0,0}, 31341} },
      {use_ebus, false},
```
xandkar, line 17: trailing comma needs to be removed now

xandkar, line 17:
```suggestion
     {use_ebus, false}
```

xandkar (approved):

PR #1716, opened 2022-06-03
Fix dialyzer warning: 6th item in request tuple is binary, not function.

ke6jjj:
Fix a dialyzer warning cause by mistakenly stating that the sixth
element of the tuple that is cached for a target request is a function().
It is not; it's a binary() containing the challenger's signature.

### DIFF
--- a/src/poc/miner_poc_grpc_client_statem.erl
+++ b/src/poc/miner_poc_grpc_client_statem.erl
@@ -869,17 +869,19 @@ process_check_target_reqs(_State = #data{self_sig_fun = SelfSigFun}) ->
         Reqs
     ).
 
+-type target_request_data() :: {
+    URI :: string(),
+    libp2p_crypto:pubkey_bin(),
+    OnionKeyHash :: binary(),
+    BlockHash :: binary(),
+    NotificationHeight :: non_neg_integer(),
+    ChallengerSig :: binary(),
+    TimestampSec :: integer()
+}.
+
 -spec cache_check_target_req(
-    ID::binary(),
-    ReqData::{
-        string(),
-        libp2p_crypto:pubkey_bin(),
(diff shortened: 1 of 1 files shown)


ke6jjj:
In case you are confused _why_ dialyzer didn't like the old type spec, the "inferred" type came from here:

https://github.com/helium/miner/blob/5027821780454880f4c9d40670f8a6c7ec0297e1/src/poc/miner_poc_grpc_client_statem.erl#L1023

The old type specified that the argument being passed as "`ChallengerSig`" was a `function()`. It is not. It's clearly a `binary()`.

Kingofglory2021:
I don’t really see an advantage from this because bobcat look scan to me after waited for 7 months I spent my time and buying with my money I believe it’s scam


Sent from the all-new AOL app for iOS


On Saturday, June 4, 2022, 12:20 am, Jeremy Cooper ***@***.***> wrote:




In case you are confused why dialyzer didn't like the old type spec, the "inferred" type came from here:

https://github.com/helium/miner/blob/5027821780454880f4c9d40670f8a6c7ec0297e1/src/poc/miner_poc_grpc_client_statem.erl#L1023

The old type specified that the argument being passed as "ChallengerSig" was a function(). It is not. It's clearly a binary().

evanmcc (approved):
This seems OK to get us clean, but ideally we'd also convert to a record here.

jeffgrunewald:
> I don’t really see an advantage from this because bobcat look scan to me after waited for 7 months I spent my time and buying with my money I believe it’s scam Sent from the all-new AOL app for iOS On Saturday, June 4, 2022, 12:20 am

Please keep the commentary relevant

jeffgrunewald (approved):

PR #1723, opened 2022-06-07
shim gateway api into ebus interface

jeffgrunewald:
The pieces of the current implementation of the miner ebus api are being deprecated one by one, including the removal of asserting locations with the change to validator-initiated challenges and now with the disabling of miners following the chain breaking the height, blockage, and related status checks.

This change temporarily grants the ebus interface a stay of execution by having it call into the embedded gateway's local grpc api to retrieve the information and allow makers additional time to transition their firmware off the ebus interface to use the `gateway_config` application.

### DIFF
--- a/src/miner_ebus.erl
+++ b/src/miner_ebus.erl
@@ -6,11 +6,15 @@
 
 -export([send_signal/2]).
 
--export([start_link/0, start_link/2, init/1, handle_message/3, handle_cast/2]).
+-export([start_link/0, start_link/2, init/1, handle_message/3, handle_cast/2, handle_info/2]).
 
--record(state, {}).
+-record(state, {
+    connection = undefined :: grpc_client:connection() | undefined,
+    grpc_api = false :: boolean()
+}).
 
 -define(SERVER, miner_ebus).
+-define(SERVICE, 'helium.local.api').
 
 -define(MINER_APPLICATION_NAME, "com.helium.Miner").
 -define(MINER_OBJECT_PATH, "/").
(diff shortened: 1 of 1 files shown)


src/miner_ebus.erl
```diff
@@ -46,52 +47,66 @@ init(_Args) ->
     erlang:register(?SERVER, self()),
     {ok, #state{}}.
 
-
-handle_message(?MINER_OBJECT(?MINER_MEMBER_ADD_GW)=Member, Msg, State=#state{}) ->
+handle_message(Member, Msg, State = #state{ connection = undefined }) ->
+    GrpcPort = application:get_env(miner, gateway_api_port, 4468),
+    case grpc_client:connect(tcp, "localhost", GrpcPort) of
+        {ok, Connection = #{http_connection := Pid}} ->
+            erlang:monitor(process, Pid),
```
madninja, line 54: you're monitoring but not handling messages related to that monitor?

src/miner_ebus.erl
```diff
@@ -46,52 +47,66 @@ init(_Args) ->
...
-        {error, Error} ->
-            lager:warning("Invalid assert_loc args: ~p", [Error]),
-            {reply_error, ?MINER_ERROR_BADARGS, Member, State}
+handle_message(?MINER_OBJECT(?MINER_MEMBER_ASSERT_LOC) = Member, _Msg, State) ->
+    lager:warning("Assert_loc deprecated; miners now obtain loc data on-chain", []),
+    {reply_error, ?MINER_ERROR_BADARGS, Member, State};
+handle_message(?MINER_OBJECT(?MINER_MEMBER_P2P_STATUS) = Member, _, State=#state{ connection = Connection }) ->
+    case call_unary(Connection, height, #{}) of
+        {ok, #{result := #{height := Height}}} ->
+            {reply, [{array, {struct, [string, string]}}], [{"connected", "yes"}, {"height", integer_to_list(Height)}], State};
```
madninja, line 89: this, and gateway_config's notion of connected may change if we do what @andymck is working towards which is throwing out the hotspot connection if it doesn't have an asserted location

src/miner_ebus.erl
jeffgrunewald: dang, good catch

madninja (approved):

PR #1724, opened 2022-06-08
Build devnet validator debs

jadeallenx:
Problem to solve: We want to package validators for devnet and publish them to packagecloud like our other validator packages.


jeffgrunewald (approved):

jeffgrunewald:
Looks like you need to add another entry to the `sed` here to cover the devnet tag prefix https://github.com/helium/miner/blob/a1fb7e0d37410fc4baa751a80abad0b6160f8e2a/.buildkite/scripts/make_deb.sh#L9

jadeallenx:
> Looks like you need to add another entry to the `sed` here to cover the devnet tag prefix

Ah you're 100% right. Thanks.

PR #1730, opened 2022-06-08
prevent tagging validator release without updating version in miner.erl

jeffgrunewald:
as long as we're tagging our releases with `tag-release.sh` and we don't plan to change the erlang versioning scheme or the tagging strategy for validator tags this should prevent us from publishing a validator release without first updating the version function in `src/miner.erl`

### DIFF
--- a/tag-release.sh
+++ b/tag-release.sh
@@ -13,16 +13,6 @@ dep_hash() {
     git show "${MINER}:rebar.lock" | grep -A2 "${DEP}" | sed -e "s/.*\"\(.*\)\".*/\1/" | tr '\n' ' ' | awk '{print $3}'
 }
 
-strip_zeros() {
-    local VZN=$(echo $1 | sed -e 's/0//g')
-
-    if [[ $VZN == "" ]]; then
-        echo "0"
-    else
-        echo $VZN
-    fi
-}
-
 val_tag() {
     local VAL_VERSION
     local MAJ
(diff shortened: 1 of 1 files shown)


jadeallenx (approved):

tag-release.sh
```diff
@@ -13,6 +13,30 @@ dep_hash() {
...
+    fi
+}
+
+val_tag() {
+    local VAL_VERSION
+    local MAJ
+    local MIN
+    local PATCH
+
+    VAL_VERSION=$(grep -A 1 "%% MMMmmmPPPP" src/miner.erl | grep -o -E '[0-9]+')
```
evanmcc, line 32: can you add a comment in `src/miner.erl` to the effect that we rely on that comment in the tagging script?

madninja:
Hell yes. Take care of Evan’s comment please?

tag-release.sh
jeffgrunewald: i can do this as well, but i tried changing the pattern to match against the structure of the erlang `version/0` function itself, so as long as no one writes any comments that end the line in a period just maintaining a proper erlang function wrapped around that version number should work

anthonyra:
I believe your `strip_zeros` function would return `0010 -> 1` a quick search suggests to use something like `echo $((10#$1))` instead

jeffgrunewald:
> I believe your `strip_zeros` function would return `0010 -> 1` a quick search suggests to use something like `echo $((10#$1))` instead

great catch; didn't test against versions with trailing zeros. thanks @anthonyra !

tag-release.sh
jeffgrunewald: any more thoughts on this vs. updating the comment to add some weight to it @evanmcc ?

Vagabond (approved):

PR #1732, opened 2022-06-09
return queue data when submitting txn

andymck:
miner side of https://github.com/helium/blockchain-core/pull/1216

Provides a new single sidecar API to submit a new txn or query a previously submitted txn as part of the new TX_PROTOCOL_V3 support.

In addition provides a set of CTs to perform end to end tests of the updated flows being exercised via the txn mgr grpc submit and query APIS


src/miner_hbbft_sidecar.erl
```diff
@@ -169,11 +181,12 @@ handle_call({submit, Txn}, From,
                                     case blockchain_txn:absorb(Txn, Chain) of
                                         ok ->
                                             spawn(fun() ->
-                                                      %% this will now return {ok, Position, Length}
-                                                      %% or {error, full} which we could feed back to the caller using gen_server:reply() on the From
-                                                      catch libp2p_group_relcast:handle_command(Group, {txn, Txn})
+                                                        %% this will now return {ok, Position, Length}
+                                                        %% or {error, full} which we could feed back to the caller using gen_server:reply() on the From
+                                                        {ok, QueuePos, QueueLen} = libp2p_group_relcast:handle_command(Group, {txn, Txn}),
```
jeffgrunewald, line 186: explicit match here seems like it could wedge the calling process since this function is being triggered by a `gen_server:call` with an infinite timeout and this spawned process is not monitored so if this call to the relcast group returns `{error, full}` and crashes on the bad match it'll just go unnoticed and the caller will continue to wait forever. Suggesting a generalized reply in my `query_txn` PR that just wraps either response in a `{_, Height}` tuple like many of the other responses returned from `handle_calls` to the sidecar here.

src/miner_hbbft_sidecar.erl
```diff
@@ -111,6 +119,10 @@ handle_call({set_group, Group}, _From, #state{group = OldGroup} = State) ->
         {P1, P2} when is_pid(P1) andalso is_pid(P2)  ->
             ok;
         {undefined, P} when is_pid(P) ->
+            ok = libp2p_swarm:add_stream_handler(blockchain_swarm:tid(), ?TX_PROTOCOL_V3,
```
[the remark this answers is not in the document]
andymck: done

src/miner_hbbft_sidecar.erl
```diff
@@ -169,11 +182,12 @@ handle_call({submit, Txn}, From,
                                     case blockchain_txn:absorb(Txn, Chain) of
                                         ok ->
                                             spawn(fun() ->
-                                                      %% this will now return {ok, Position, Length}
-                                                      %% or {error, full} which we could feed back to the caller using gen_server:reply() on the From
-                                                      catch libp2p_group_relcast:handle_command(Group, {txn, Txn})
+                                                        %% this will now return {ok, Position, Length} or {error, full}
+                                                        %% which we feed back to the caller using gen_server:reply() on the From
+                                                        Reply = libp2p_group_relcast:handle_command(Group, {txn, Txn}),
```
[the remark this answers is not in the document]
andymck: I believe this is ok

if the process crashes, the dialer will timeout and ultimately be retried.  the catch was providing no additional value here other than hiding a crash reason and leading to  an ok response being automatically returned to the caller (even tho the txn may have failed ).  

Now the caller will get a representative response and all callers from what I can see use their own catch around calls to the sidecar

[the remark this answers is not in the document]
jeffgrunewald: it's been in a spawned anonymous process all along so this actually seems somewhat better in that it'll surface failed submissions faster than returning the automatic `ok` it was previously returning. if timeouts are set really long and we're worried about them slowing down the system we could always try and catch again as long as we send the error back in the out-of-band reply

src/miner_hbbft_sidecar.erl
```diff
@@ -85,6 +86,13 @@ new_round(Buf, BinTxns) ->
 prefilter_round(Buf, Txns) ->
     gen_server:call(?SERVER, {prefilter_round, Buf, Txns}, infinity).
 
+handle_txn(submit = RequestType, Txn) ->
+    lager:debug("handling txn with request type ~p", [RequestType]),
+    gen_server:call(?SERVER, {submit, Txn}, infinity);
+handle_txn(update = RequestType, Txn) ->
+    lager:debug("handling txn with request type ~p", [RequestType]),
```
jeffgrunewald, line 93: these two debug log statements look like they're just extra noise in context of the debug statements we're also logging in their accompanying `handle_call/3` functions; would propose consolidating those into a single statement in the handler.

test/miner_txn_mgr_SUITE.erl
```diff
@@ -358,7 +366,433 @@ txn_dependent_test(Config) ->
...
+    %% which in this case will be a single consensus member
+    %% this will result in the txn mgr pushing its cached txns
+    %% to this single CG member
+    %% and thus we can ensure we only need to check the sidecar
+    %% buffer on a single and known CG member
+
+    %% also meck out is_valid for payment txns
+    %% this test is not concerned with validity issues
+    meck:new(blockchain_txn_mgr, [passthrough]),
+    meck:new(blockchain_txn_payment_v1, [passthrough]),
```
jeffgrunewald, line 682: any reason to suspect a v2 payment wouldn't work just as well here for validation purposes if v1 is deprecated (it is, isn't it?) other than creating needing to create the payments separately from the txn?

test/miner_txn_mgr_SUITE.erl
```diff
@@ -358,7 +366,433 @@ txn_dependent_test(Config) ->
...
+    PayeeAddr = miner_ct_utils:node2addr(Payee, AddrList),
+    {ok, _Pubkey, SigFun, _ECDHFun} = ct_rpc:call(Miner, blockchain_swarm, keys, []),
+    IgnoredTxns = [],
+
+    %% meck out the sidecar random_n function
+    %% make it return an empty list
+    %% simulate an empty buffer
+    %% this should result in the submitted txns
+    %% remaining in the sidecar buffer
+    meck:new(hbbft_utils, [passthrough]),
```
jeffgrunewald, line 657: i'm surprised you can tell the remote rpc nodes to expect a meck intercept of a function call when you've only created the meck process on the test running node/process

src/miner_hbbft_sidecar.erl
andymck: done

test/miner_txn_mgr_SUITE.erl
andymck: i too was surprised but thankful

test/miner_txn_mgr_SUITE.erl
andymck: any type of txn with a nonce check dependency will work here.  v1 payments were used as that is what was already in the existing mgr tests

evanmcc (approved):

PR #1735, opened 2022-06-09
DNM!! Configure seed nodes to use a network id

jadeallenx:
Configure seeds to use a network id. Also enable devnet_seed packages

TODO: Update/bump libp2p and blockchain-core branches and rebar.lock


jadeallenx:
Took these commits and merged into the seed-build branch

PR #1736, opened 2022-06-09
Make the smallest bloom filter 1000 elements

jadeallenx:
Problem to solve: On new networks, the actual gateway count can be 0 which causes a crash in miner that prevents blocks from syncing.

Solution: Make the size of the new filter at least 1000, and if it's larger, use the actual gateway count for filter sizing.


evanmcc (approved):

JayKickliter (approved):
Looks good, and thanks for the descriptive comments.

For other reviwers: there's an upstream erbloom [PR](https://github.com/Vagabond/erbloom/pull/2) to not panic on size==0, but even with that fix we'd still get a badmatch here, so the fix in this PR seems like a good idea.

PR #1738, opened 2022-06-10
handle query_txn requests

jeffgrunewald:
Adds a handler function for the `query_txn` / `update` transaction type that is now possible via the hbbft_sidecar's `handle_txn/2` api function.

### DIFF
--- a/src/handlers/miner_hbbft_handler.erl
+++ b/src/handlers/miner_hbbft_handler.erl
@@ -246,6 +246,15 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
                     {reply, {ok, Position, Length}, fixup_msgs(Msgs), State#state{hbbft=NewHBBFT}}
             end
     end;
+handle_command({query_txn, Txn}, State = #state{hbbft = HBBFT}) ->
+    Buf = hbbft:buf(HBBFT),
+    SerTxn = blockchain_txn:serialize(Txn),
+    case current_buffer_position(SerTxn, Buf) of
+        {ok, Position} when is_integer(Position) ->
+            {reply, {ok, Position, length(Buf)}, ignore};
+        {error, not_found} ->
+            {reply, {error, not_found}, ignore}
+    end;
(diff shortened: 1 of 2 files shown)


src/handlers/miner_hbbft_handler.erl
```diff
@@ -246,6 +246,15 @@ handle_command({txn, Txn}, State=#state{hbbft=HBBFT}) ->
                     {reply, {ok, Position, Length}, fixup_msgs(Msgs), State#state{hbbft=NewHBBFT}}
             end
     end;
+handle_command({query_txn, Txn}, State = #state{hbbft = HBBFT}) ->
+    Buf = hbbft:buf(HBBFT),
+    SerTxn = blockchain_txn:serialize(Txn),
+    case current_buffer_position(SerTxn, Buf) of
```
andymck, line 252: why not use lists:member here?

andymck: nvmd, you are avoiding having to iterate twice over the list

src/miner_hbbft_sidecar.erl
```diff
@@ -181,10 +181,10 @@ handle_call({submit, Txn}, From,
...
                                         ok ->
                                             spawn(fun() ->
-                                                        %% this will now return {ok, Position, Length}
-                                                        %% or {error, full} which we could feed back to the caller using gen_server:reply() on the From
-                                                        {ok, QueuePos, QueueLen} = libp2p_group_relcast:handle_command(Group, {txn, Txn}),
-                                                        gen_server:reply(From, {ok, QueuePos, QueueLen, Height})
+                                                        %% this will now return {ok, Position, Length} or {error, full}
+                                                        %% which we feed back to the caller using gen_server:reply() on the From
+                                                        Reply = libp2p_group_relcast:handle_command(Group, {txn, Txn}),
+                                                        gen_server:reply(From, {Reply, Height})
```
andymck, line 187: I note you have changed the tuple format of the response here with the height outside of the {ok, ...} tuple.  No objections to that but i believe there are a couple of other replys that will need updated to maintain consistency

PR #1742, opened 2022-06-13
Add denylist to cli and jsonrpc

PaulVMo:
Based on a request from the community, this change adds basic denylist functionality to the cli and jsonrpc endpoint. Supported methods are:
- `denylist status` which returns whether a denylist is loaded and if so which version
- `denylist check <address>` which returns whether or not the provided hotspot address is on the loaded list; will return true/false only if denylist is loaded and error if the address is not a valid b58 hotspot address

### DIFF
--- a/src/cli/miner_cli_denylist.erl
+++ b/src/cli/miner_cli_denylist.erl
@@ -1,0 +1,106 @@
+%%%-------------------------------------------------------------------
+%% @doc miner_cli_denylist
+%% @end
+%%%-------------------------------------------------------------------
+-module(miner_cli_denylist).
+
+-behavior(clique_handler).
+
+-export([register_cli/0]).
+
+register_cli() ->
+    register_all_usage(),
+    register_all_cmds().
+
+register_all_usage() ->
+    lists:foreach(fun(Args) ->
+                          apply(clique, register_usage, Args)
+                  end,
+                 [
+                  denylist_usage(),
+                  denylist_status_usage(),
+                  denylist_check_usage()
+                 ]).
+
+register_all_cmds() ->
+    lists:foreach(fun(Cmds) ->
+                          [apply(clique, register_command, Cmd) || Cmd <- Cmds]
+                  end,
+                 [
+                  denylist_cmd(),
+                  denylist_status_cmd(),
+                  denylist_check_cmd()
+                 ]).
+%%
+%% denylist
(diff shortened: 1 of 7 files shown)


src/cli/miner_cli_denylist.erl
```diff
@@ -0,0 +1,106 @@
...
+denylist_check_usage() ->
+    [["denylist", "check"],
+     ["denylist check <address>\n\n",
+      "  Check if address is on denylist.\n\n"
+     ]
+    ].
+
+denylist_check(["denylist", "check", Address], [], []) ->
+    Result =
+        try miner_poc_denylist:check(libp2p_crypto:b58_to_bin(Address)) of
```
jeffgrunewald, line 98: as a user-facing function, i'm not sure i like returning false here when the denylist isnt loaded or configured since it could give the false impression that they're not on the denylist when they actually are. without doing more changes to adjust that underlying behavior and catch instances of the check function returning `not_configured` instead of `false` we could just put the check in the cli behind a call to the status function?

from an internal standpoint i think it's cleaner for the denylist to always return false since the effect is the same - to allow the poc to proceed uninterrupted but here i think it has more implications since the outcome could determine the user's next troubleshooting steps.

src/cli/miner_cli_denylist.erl
```diff
@@ -0,0 +1,106 @@
...
+      "  Display status of denylist.\n\n"
+     ]
+    ].
+
+denylist_status(["denylist", "status"], [], []) ->
+    Text = try miner_poc_denylist:get_version() of 
+        {ok, Version} -> 
+            clique_status:text(io_lib:format("Denylist version ~p loaded", [Version]))
+    catch _:_ ->
+        clique_status:text(io_lib:format("No denylist loaded",[]))
```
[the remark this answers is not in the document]
PaulVMo: I expanded on the message to give more detail and also added a separate case for missing config versus failing to load for another reason

src/cli/miner_cli_denylist.erl
PaulVMo: Agree. I think i was avoiding the extra complication of the extra check but have now added it in. Let me know what you think.

jeffgrunewald (approved):

jeffgrunewald:
Looks good!

PR #1759, opened 2022-06-28
Disable cache prewarm on miners

JayKickliter:
This can't go in as-is since it will affect validators. I'm open to suggestions to get to the end goal of only disabling prewarm on miners.


evanmcc:
This probably also needs a mode check in the vars fun too.

joecaswell:
Since it is already set to false by default in https://github.com/helium/miner/blob/master/config/val.config.src#L26, wouldn't this change only impact nodes based on the miner build?

PR #1766, opened 2022-07-05
Force validator selection for testing purposes

jadeallenx:
Problem: It is sometimes useful to force a hotspot to choose a very specific validator for gRPC connections

Solution: Make the validator selection significantly less random.

([xkcd on random numbers][1])

[1]: https://xkcd.com/221/


JayKickliter:
Testing is done, explicitly closing without merging.

PR #1771, opened 2022-07-07
Let relcast do term serialization, if needed

Vagabond:
Relates to helium/relcast#50


src/handlers/miner_dkg_handler.erl
```diff
@@ -312,8 +312,7 @@ serialize(State) ->
            sent_conf => SentConf,
            height => Height,
            delay => Delay},
-    M = maps:map(fun(_K, Term) -> t2b(Term) end, M0),
-    maps:merge(PreSer, M).
+    maps:merge(PreSer, M0).
```
marcsugiyama, line 315: PreSer is an empty map. Do we need the `maps:merge`?

src/handlers/miner_dkg_handler.erl
```diff
@@ -322,7 +321,7 @@ deserialize(MapState0) when is_map(MapState0) ->
     MapState = maps:map(fun(_K, undefined) ->
                                 undefined;
                            (_K, B) ->
-                                binary_to_term(B)
+                                B
```
marcsugiyama, line 324: I think this `maps:map` is now a no-op?

src/handlers/miner_hbbft_handler.erl
```diff
@@ -463,7 +462,7 @@ deserialize(#{sk := SKSer,
                   case StateMap of
                       #{K := V} when V /= undefined andalso
                                      V /= Bundef ->
-                          binary_to_term(V);
+                          V;
                       _ when K == sig_phase ->
                           sig;
                       _ when K == seen ->
```
marcsugiyama, line 468: Since we're ignoring these fields in `deserialize` do we need to serialize them in the first place? Skipping these fields in `serialize` will reduce the number of field-by-field compares in `relcast`.

marcsugiyama (approved):

src/handlers/miner_dkg_handler.erl
Vagabond: No, good catch.

src/handlers/miner_dkg_handler.erl
Vagabond: yes, will fix.

src/handlers/miner_hbbft_handler.erl
Vagabond: I think its populating the default values if they're not defined.

marcsugiyama: OK, I get it now. The first case clause checks for `undefined` and falls through to the others if it is.

evanmcc (approved):

Issue #1083, opened 2021-09-24
Regression around the management of rssi


BenoitDuffez:
I suspect there is a regression in https://github.com/helium/miner/commit/317f1edf089ec5d17c673016a527be6f9d7f5dd1#diff-72653b0c74b5401df251b16d8460def13ffde2b484db58696c596c5ec4898bfaL642-R648 is where the `rssis`/`rssic` split has been modified a bit, and one invocation of `miner_lora:packet_rssi` uses a static `true` for `UseRSSIS` where I think it should use

    UseRSSIS = case Chain /= undefined andalso blockchain:config(?poc_version, blockchain:ledger(Chain)) of
        {ok, X} when X > 10 -> true;
        _ -> false
    end,

like it does in another place in the file.

BenoitDuffez:
Also, I suspect we should not use `rssis` but rather `rssic`.

Vagabond:
It's a static true in the other case because that's for normal packets, and those don't affect consensus.

RssiC is Channel RSSI, RSSIS is signal RSSI as I understand it.

BenoitDuffez:
We don't send the RSSI in `rssis` but rather in `rssic`. We always have been, and packet forwarding worked a few weeks ago.

Now I see a crash:

```
2021-09-24 12:29:07.681 1528 [info] <0.1578.0>@miner_lora:handle_udp_packet:427 PUSH_DATA [{<<"rxpk">>,[[{<<"aesk">>,0},{<<"brd">>,261},{<<"codr">>,<<"4/5">>},{<<"data">>,<<"AF9ZUlZJXzQ5bmVnQaRPAwA5Oy2R4cw=">>},{<<"datr">>,<<"SF7BW125">>},{<<"freq">>,868.1},{<<"jver">>,2},{<<"modu">>,<<"LORA">>},{<<"rsig">>,[[{<<"ant">>,0},{<<"chan">>,5},{<<"lsnr">>,-7.0},{<<"rssic">>,-115}]]},{<<"size">>,23},{<<"stat">>,1},{<<"time">>,<<"2021-09-24T12:29:07.665049Z">>},{<<"tmst">>,421046188}]]}] from 8248060143997291342 on 43682
2021-09-24 12:29:07.683 1528 [notice] <0.1578.0>@miner_lora:handle_packets:600 Routing {eui,931991820658030,4122024327536335199}
[...]

PR #1777, opened 2022-07-21
fix rssi computation from rssic if SNR < 0

BenoitDuffez:
rssis is supposed to be the useful signal strength
rssic is supposed to be the received signal strength

The blockchain expects the rssi to be the rssis. However when falling back to rssic when rssis is unavailable, the computation is wrong when SNR < 0
Indeed, in that case the useful signal strength is below the noise level, which is rssic. So the rssis = SNR + rssic when SNR < 0
This should help use the appropriate GWMPv2 value from the packet forwarders

### DIFF
--- a/src/miner_lora.erl
+++ b/src/miner_lora.erl
@@ -842,7 +842,11 @@ packet_rssi(Packet, UseRSSIS) ->
                 false ->
                     %% Just use RSSIC.
                     fun (Obj) ->
-                        maps:get(<<"rssic">>, Obj, undefined)
+                        SNR = maps:get(<<"lsnr">>, Obj, 0),
+                        case SNR < 0 of
+                            true -> SNR + maps:get(<<"rssic">>, Obj, undefined);
+                            _ -> maps:get(<<"rssic">>, Obj, undefined)
+                        end
                     end
             end,
             BestRSSISelector =


Vagabond:
When is RSSIS unavailable on GWMPv2?

BenoitDuffez:
rssic/rssis are available on RF implementations of Semtech reference design v2.x  
rssic only (rssis unavailable) is on RF implementations of Semtech reference design v1.x (most if not all of Helium hotspots)

BenoitDuffez:
I think this is the continuation of this: https://github.com/helium/miner/issues/1083#issuecomment-926611790

Vagabond:
Where did you find `rssis = SNR + rssic when SNR < 0`? I've not seen that before

BenoitDuffez:
When SNR < 0, the signal is below the noise level, hence rssic is the noise level. This, signal level is rssic+snr

pdvaneijk:
Just to confirm, so for any Helium gateway that either uses the SX1301/08 or SX1302/03:

if SNR is positive, use rssic

if snr<0, packet power is rssi+snr

The Semtech V2.X gateway is a totally different beast (ADI RF front end, large FPGA + 2-8 SX1301s) and there are no Helium gateways based on the design.

BenoitDuffez:
I might get a pit pedantic, but my PR is about JSON format (GWMP version) and not RF ref des. Ref des v1 can use GWMPv2 (all Kerlink do).

ke6jjj:
For context, everyone, the point here is to pull out the best possible estimation of the received signal power so that the Free-Space Path Loss equation can accurately rule on signals that are impossibly strong for the distance asserted.

pdvaneijk:
> I might get a pit pedantic, but my PR is about JSON format (GWMP version) and not RF ref des. Ref des v1 can use GWMPv2 (all Kerlink do).
> 
> Also, on a side note, I'm using a DIY miner with a Kerlink ref des v2 gateway (the 2x1301 arch you mentioned), it works fine (besides the SNR < 0 issue mentioned here). So there's at least one! 😁

Semtech V2.X gateways use V2 GWMP and Semtech V1 Gateway and Corecell use V1 GWMP so I think we are good

BenoitDuffez:
If you use the semtech packet forwarder, yes. Kerlink hotspots do not, they all use GWMPv2. So this PR is needed.

ke6jjj (approved):

src/miner_lora.erl
```diff
@@ -842,7 +842,11 @@ packet_rssi(Packet, UseRSSIS) ->
                 false ->
                     %% Just use RSSIC.
                     fun (Obj) ->
-                        maps:get(<<"rssic">>, Obj, undefined)
+                        SNR = maps:get(<<"lsnr">>, Obj, 0),
+                        case SNR < 0 of
+                            true -> SNR + maps:get(<<"rssic">>, Obj, undefined);
```
jeffgrunewald, line 847: Couldn’t this line result in an arithmetic operation error if rssic is undefined?

ke6jjj: Looks like it would, yes. I think a missing `rssic` is a brutal fatal error, though. So the arithmetic error is appropriate to propagate?

Vagabond: it's defined to be present in this protocol version

PR #1780, opened 2022-07-25
Use a different Dockerfile for validator builds

Vagabond:
Less disruptive version of  #1488


Dockerfile-ubuntu
```diff
@@ -0,0 +1,105 @@
...
+       libsodium-dev \
+       libssl-dev \
+       libtool \
+       lz4 \
+       libprotobuf-dev \
+       wget \
+       ${EXTRA_BUILD_APT_PACKAGES} 
+
+# Strip debug symbols from ERTS binarys to reduce size
+RUN scanelf --nobanner -E ET_DYN -BF '%F' --recursive /usr/local/lib/erlang | xargs -r strip --strip-unneeded
```
jeffgrunewald, line 36: is this as effective as we'd like if it's run before compiling the application?

evanmcc (approved):

PR #1790, opened 2022-08-05
ara/deterministic-timing-rework

anthonyra:
I needed to add a .devcontainer since I use an M1 Mac. Somehow, vscode also added *.log files which I'll need to figure out how to purge before an official PR is submitted.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,22 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.159.0/containers/javascript-node
+{
+	"name": "Erlang",
+	"build": {
+		"dockerfile": "remote.Dockerfile",
+		"args": { "VARIANT": "24.3.4.0-alpine" }
+	},
+
+	// Set *default* container specific settings.json values on container create.
+	"settings": { 
+		"terminal.integrated.shell.linux": "/bin/sh"
+	},
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"pgourlain.erlang"
+	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [3000],
+}
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -20,9 +20,11 @@
     keys/0,
 
     reset_late_block_timer/0,
+    calculate_next_block_time/1, calculate_next_block_time/2,
 
     start_chain/2,
     install_consensus/1,
+    group_block_time/1,
     remove_consensus/0,
     version/0
 ]).
@@ -55,8 +57,18 @@
         snapshot_hash  => binary()
      }.
 
+-type metadata_v3() ::
+    #{
+        timestamp           => integer(),
+        seen                => binary(),
+        bba_completion      => binary(),
+        head_hash           => blockchain_block:hash(),
+        snapshot_hash       => binary(),
+        target_block_time   => integer()
+     }.
+
 -type metadata() ::
-    [{J :: pos_integer(), M :: metadata_v2() | metadata_v1()}].
+    [{J :: pos_integer(), M :: metadata_v3() | metadata_v2() | metadata_v1()}].
 
 -type swarm_keys() ::
     {libp2p_crypto:pubkey(), libp2p_crypto:sig_fun()}.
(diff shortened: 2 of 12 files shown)


anthonyra:
Some added context, the attempt here is to include a lazy vote in regards to the next block time for CG members. It utilizes BBA metadata to do this. In doing so it's always voting for the "next" block.

Things to consider
- Since the next block is being voted on, it defaults to old calculations for block time and upgrades to the group time if applicable
- I also found that the old calculation lead to some very "annoying" behavior depending block history. As the block history of the validator increases it becomes significantly less sensitive to current block time concerns. An example, if the validator has 50k blocks of history and the average is 60.2 seconds when a block is added that's 80 seconds the response is "weak". In comparison, if it had a block history of only 300 and 60.1 seconds average that 80 second block will result in a much "faster" block time.

In response to the second bullet, I "hacked" together and upper and lower bound for block history. The upper bound is greater then snapshot size but less then 50k. The lower bound is snapshot size. It then performs a similar calculation as before but chooses the most aggressive "slow" average for block times. It also tries to be the most aggressive in the beginning of an epoch then tappers down closer to target block time. This is to ensure that all nodes are as close to the same level of "work" when approaching those thicc reward blocks.

TODOS:
- [ ] Test with a group that has different block time histories

src/miner.erl
```diff
@@ -750,75 +801,166 @@ common_enough_or_default(Threshold, Xs, Default) ->
...
+
+
+%% ------------------------------------------------------------------
+%% EUNIT Tests
+%% ------------------------------------------------------------------
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+
+%% Confirm changes to make catchup_time proportional
+catchup_time_test() ->
```
PaulVMo, line 955: these unit test where specific to my changes in #1557 . Remove if not using that logic or they will not work.

src/miner.erl
```diff
@@ -750,75 +801,166 @@ common_enough_or_default(Threshold, Xs, Default) ->
...
-            true ->
-                case blockchain:find_first_block_after(StartHeight, Chain) of
-                    {ok, Actual, StartBlock} ->
-                        {Actual, blockchain_block:time(StartBlock)};
-                    _ ->
-                        {0, undefined}
-                end;
-            false ->
-                {0, undefined}
+    %% mimic snapshot_take functionality for block range window
```
PaulVMo, line 822: What's the benefit of looking at the very short ("snapshot_take") block time average here? I fear this is going to add a lot of volatility to block times (in addition to adding complexity to this code). I believe you are trying to catch up quickly. However, to what benefit? I think it would be preferable to catch up over a longer period of time to have less volatility in times from block to block. There is a high chance that this period includes multiple rewards blocks which the algorithm will try to quick compensate for which exceptionally quick in my opinion is unnecessary.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
```
PaulVMo, line 1: can't you add this file to .gitignore or something? or not add it to your commit?

PaulVMo:
Good stuff Anthony. Thanks for working on this. This issue has plagued the community for too long and they (myself included) will be happy to see it abolished. Gone will be the days of so meticulously worrying over how much block history is on a validator.

I did add some review comments. I am not sold on the changes to make reactions "faster". We are talking about timeframes of 30 days / ~50K blocks to make up for lost blocks / emissions, I do not think we need to make that all up so quickly. Especially with rewards being several minutes. I think we will end up seeing a long reward block, followed by five to ten 50 second blocks and then 20 or so 60s blocks which then repeats. That puts undo pressure on validators to catch up so quickly all the time. I think we'd be better off with the long window and letting validators catch up over the whole epoch (or multiple epoch) say with a long reward block followed by 30 block at at 57 or 58 seconds rather than slamming on the gas for a few blocks and then backing off.

src/miner.erl
anthonyra: Oh yea, haven't gotten to "tests" yet. But figured I'd post this to get some more eyes on it.

src/miner.erl
```diff
@@ -750,75 +801,166 @@ common_enough_or_default(Threshold, Xs, Default) ->
...
 %% when drift is small or 0, let it accumulate for a bit
 catchup_time(N) when N < 0.001 ->
     0;
-%% when it's still relatively small, apply gentle adjustments
-catchup_time(N) when N < 0.01 ->
-    1;
-%% if it's large, jam on the gas
-catchup_time(_) ->
-    10.
+%% try and catch up within 10 blocks, max 10 seconds
```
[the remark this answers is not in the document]
anthonyra: After some forced halts and restarts. The proportional and original catch-up calculations were far more aggressive then what I proposed. I guess I forgot to mention, that it still tries to reach 60 seconds over the course of an epoch. However, the quicker block times occur early in the epoch compared to later. It also leads to less bouncing due to the "hunting mechanisms" sensitivity.

PR #1800, opened 2022-08-16
Fix some memory leaks for validators

Vagabond:


src/miner.erl
```diff
@@ -686,6 +686,7 @@ select_transactions(Chain, Txns, {ElectionEpoch0, EpochStart0}, BlockHeightCurr,
                          _       -> blockchain_txn_rewards_v1
                      end,
                 {ok, Rewards} = RewardsMod:calculate_rewards(Start, End, Chain),
+                blockchain_hex:destroy_memoization(),
```
evanmcc, line 689: I'm not sure how this got deleted!  I wonder what the impact in ETL will be.

evanmcc (approved):

PR #1801, opened 2022-08-16
Disable p2p connections when the node is not running the chain

Vagabond:


Vagabond:
This appears to be working well in real-world testing.

evanmcc (approved):

src/miner_critical_sup.erl
```diff
@@ -77,7 +77,17 @@ init(_Opts) ->
     case {MinerMode, GatewaysRunChain} of
         {gateway, false} ->
             lager:info("grpc gateway, not loading chain"),
-            application:set_env(blockchain, autoload, false);
+            application:set_env(blockchain, autoload, false),
+            case application:get_env(miner, gateway_and_mux_enable) of
```
andymck, line 81: why not use the 'mode' var here, which for gateways will be set to 'gateway' ?

PR #1804, opened 2022-08-17
add ability to not gossip with miners

evanmcc:
since the miner stopped following the chain, not connecting to them blackholes a gossip link, making it hard for certain nodes to connect to active chain followers and stay synced.  this PR makes prefer non-miners as gossip peers, hopefully improving sync until we can remove the miners from the gossip network, at which point we can tune down the number of active connections.

### DIFF
--- a/src/miner_util.erl
+++ b/src/miner_util.erl
@@ -15,6 +15,7 @@
          metadata_fun/0,
          random_val_predicate/1,
          random_miner_predicate/1,
+         random_non_miner_predicate/1,
          true_predicate/1,
          has_valid_local_capability/2,
          hbbft_perf/0,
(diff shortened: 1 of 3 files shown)


src/miner_util.erl
```diff
@@ -144,6 +145,10 @@ random_miner_predicate(Peer) ->
     not libp2p_peer:is_stale(Peer, timer:minutes(360)) andalso
         maps:get(<<"release_info">>, libp2p_peer:signed_metadata(Peer), undefined) /= undefined.
 
+random_non_miner_predicate(Peer) ->
+    not libp2p_peer:is_stale(Peer, timer:minutes(360)) andalso
+        maps:get(<<"release_info">>, libp2p_peer:signed_metadata(Peer), undefined) == undefined.
```
jeffgrunewald, line 150: guessing this only gets set in the validator release so a miner is always going to return the default value supplied

andymck:
Would using <<"node_type">> not be a better option to check here ? Which would have a value of <<"gateway">> for miners ?

jadeallenx (approved):

Vagabond (approved):

PR #1807, opened 2022-08-18
Provide default in code for hotspot denylist

mbthiery:
Some hotspots on the denylist seem to continue to be part of POC. There seems to be no reason that this could provide a meaningful performance advantage to validators, so I think it must be accidental.

I believe this could be due to some operators running a custom `sys.config.src` overlay that does not define the denylist. This PR puts a default in the code, forcing the operator to place "none" into the denylist_url intentionally.


Vagabond:
You probably need to do this for `denylist_keys` as well, at a minimum?

evanmcc (approved):
seems reasonable to me.

mbthiery:
@vagabond good catch. does this handle it?

src/poc/miner_poc_denylist.erl
```diff
@@ -18,14 +18,18 @@ start_link() ->
...
+                    gen_server:start_link({local, ?MODULE}, ?MODULE, [github_release, "https://api.github.com/repos/helium/denylist/releases/latest", DenyKeys], []);
                 DenyURL ->
                     Type = application:get_env(miner, denylist_type, github_release),
                     gen_server:start_link({local, ?MODULE}, ?MODULE, [Type, DenyURL, DenyKeys], [])
             end;
-        undefined ->
+        "none" ->
             ignore;
+        undefined ->
+            gen_server:start_link({local, ?MODULE}, ?MODULE, [github_release, "https://api.github.com/repos/helium/denylist/releases/latest", []], []);
```
Vagabond, line 32: I think you need to supply the default denylist keys here?

Vagabond:
A perhaps easier way to do this is to put the vars in the miner.app.src file env section like this:

https://github.com/erlang-lager/lager/blob/master/src/lager.app.src#L15

Then the config.sys file overrides the defaults from the .app file, if present.

mbthiery:
@Vagabond is my latest update what you had in mind regarding the `miner.app.src` env?

src/miner.app.src
```diff
@@ -36,7 +36,17 @@
     grpc_client
    ]},
   {included_applications, [blockchain, sibyl]},
-  {env,[]},
+  {env,
```
abhay, line 39: add a comment here that says it needs to be updated in the configs as well? and a comment in the configs to point here?

mbthiery:
@abhay Added some comments on denylist configs

abhay (commented):
LGTM. @Vagabond ?

Vagabond:
Yes, I think this is right

Vagabond (approved):

PR #1808, opened 2022-08-19
Fix jsonrpc ledger balance for entry v2

PaulVMo:
Update the JSONRPC call for account balances to use ledger entry v2 when the token_version is set to 2 (or higher). This allows the JSONRPC call to continue to pull current balance since the MOBILE / subnetworks went were activated.

### DIFF
--- a/src/jsonrpc/miner_jsonrpc_ledger.erl
+++ b/src/jsonrpc/miner_jsonrpc_ledger.erl
@@ -12,20 +12,22 @@
 
 handle_rpc(<<"ledger_balance">>, []) ->
     %% get all
+    Ledger = get_ledger(),
     Entries = maps:filter(
         fun(K, _V) ->
             is_binary(K)
         end,
-        blockchain_ledger_v1:entries(get_ledger())
+        blockchain_ledger_v1:entries(Ledger)
     ),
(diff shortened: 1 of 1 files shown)


src/jsonrpc/miner_jsonrpc_ledger.erl
```diff
@@ -166,12 +168,21 @@ get_ledger() ->
...
     end.
 
-format_ledger_balance(Addr, Entry) ->
-    #{
-        <<"address">> => ?BIN_TO_B58(Addr),
-        <<"nonce">> => blockchain_ledger_entry_v1:nonce(Entry),
-        <<"balance">> => blockchain_ledger_entry_v1:balance(Entry)
-    }.
+format_ledger_balance(Addr, Entry, Ledger) ->
+    case blockchain_ledger_v1:config(token_version, Ledger) of
```
vihu, line 172: You could call `blockchain_ledger_v1:versioned_entry_mod_and_entries_cf/1` and not have to do the token_version check here by hand. So this function can be reduced to something like this:

```erlang
{EntryMod, _} = blockchain_ledger_v1:versioned_entry_mod_and_entries_cf(Ledger),
#{<<"nonce">> = EntryMod:nonce(Entry), ...} 
```

Vagabond (approved):

PR #1822, opened 2022-10-04
Upgrade poc_mgr records before signing the receipts txn

Vagabond:


src/poc/miner_poc_mgr.erl
```diff
@@ -718,7 +718,7 @@ submit_receipts(
                     %% take only the limit
                     Witnesses2 = lists:sublist(Witnesses1, PerHopMaxWitnesses),
                     E = blockchain_poc_path_element_v1:new(Challengee, Receipt, Witnesses2),
-                    [E | Acc]
+                    [blockchain_poc_path_element_v1:maybe_upgrade(E) | Acc]
```
andymck, line 721: Is this required now?  Line 720 above is inititalizing the path the the upgraded receipt and witnesses from local_pocs/1 and so this subsequent upgrade should be redundant.

andymck (approved):

jeffgrunewald (approved):

PR #1831, opened 2022-10-07
[DNM] Adt/beacon testing

JayKickliter:
This is reference-only PR for some testing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ FROM ${RUNNER_IMAGE} as runner
 ARG VERSION
 ARG EXTRA_RUNNER_APK_PACKAGES
 
-RUN apk add --no-cache --update ncurses dbus libsodium libstdc++ \
+RUN apk add --no-cache --update ncurses dbus libsodium libstdc++ curl \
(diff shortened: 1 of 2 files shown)


config/val.config.src
```diff
@@ -12,6 +12,7 @@
   ]},
  {blockchain,
   [
+   {aux_ledger_dir, "aux"},
```
jeffgrunewald, line 15: This is on the validator config, did you want it on a config used by miner?

priv/gateway_rs/settings.toml
```diff
@@ -1,2 +1,8 @@
 ## api listening port override
 api = 4468
+
+[poc]
+entropy_uri = "https://entropy.helium.io/entropy"
+ingest_uri = "http://mainnet-pociot.helium.io:9080"
+beacon_interval = 600
```
jeffgrunewald, line 7: beacon_interval seems to be the only setting you actually need here. Ingest token is not needed on the lora network ingestor and the two URIs match the default values. 
you could set this as an environment variable for the gateway in miner_gateway_port as `GW_POC_BEACON_INTERVAL=600` if you prefer but I thought we had to rename that to just `interval`?

PR #1835, opened 2022-10-24
Confirm there are GRPC subscribers before generating PoC keys in heartbeat

PaulVMo:
This change is indented to prevent poorly configured validators from proposing PoC challenges if they will not be able to receive receipts for those challenges due to their gRPC port being inaccessible. This will enable more of the selected PoC to succeed as validators with the GRPC port blocked will never generate challenges in the first place.

The idea is for the validator to check if it has any GRPC subscribers (i.e. connected hotspots) and only if they do, generate ephemeral PoC keys for including in heartbeat txns. I do this by looking at whether there are any members in the process group for the notifications that hotspots subscribe to. I am open to other ideas for how to do this check. I was not familiar enough with Sibyl and GRPCBOX to figure out if there was an easier/better way for getting a count of connections, streams, or sessions from one of them.

Validators will still heartbeat and will still be eligible for the consensus if gRPC is not open. However, they will miss out on PoC reward and no longer contribute to a lower PoC completion rate.

Note, this is a "self-policing" type approach. Someone could workaround this check by modifying the code and there is no consensus mechanism to prevent this. However, there is precedent for this. The check that the validator is not relayed before generating a heartbeat is in the same situation. However, that has shown to be quite effective because most inaccessible ports are the result of misconfiguration rather than malicious intent.


src/miner_val_heartbeat.erl
```diff
@@ -105,7 +107,15 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
                                     %% key proposals for this heartbeat
                                     %% hashes of the public keys are included in the HB
                                     %% public and private keys are cached locally
-                                    {EmpKeys, EmpKeyHashes} = generate_poc_keys(Ledger),
+                                    %% but only do this is there are gRPC clients which indicates
```
andymck, line 110:
```suggestion
                                    %% but only do this if there are gRPC clients which indicates
```

andymck (approved):
I think this is fine as long as the limitations are understood.

src/miner_val_heartbeat.erl
```diff
@@ -105,7 +107,15 @@ handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}},
...
+                                    %% that this validator is configured to receive poc receipts
+                                    {EmpKeys, EmpKeyHashes} = case ?BYPASS_GRPC_CHECK orelse length(pg:get_members(<<"activity_check_notification">>)) > 0 of
```
andymck, line 112: for completeness you could expose an API in sibyl_bus "get_members/1".  I dont think we need to support erlang versions < 23 anymore but if we do adding this API and support both PG and PG2 would def be appropriate.

PaulVMo: Thanks. I added the function to sibyl_bus as suggested (https://github.com/helium/sibyl/pull/75) and pushed an update to miner_val_heartbeat to use that function.

PaulVMo:
This may not work, at least not how I imaged. The key proposal length is validated to be exactly the target length, versus <= as I had assumed. So this would cause the heartbeat to fail entirely. That may be a valid behavior although different than what I was originally envisioning / proposed, and a bit more extreme of a penalty for not have GRPC open. 


https://github.com/helium/blockchain-core/blob/a34308a576a40594d5057d29570f419516dc87b1/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl#L171

PR #1839, opened 2022-10-26
add exp backoff to grpc client reconnects to vals

andymck:
Adds exponential backoff to reconnects to validators.  The reconnects will start at 5 sec intervals and with each attempt expodentially increase up to a max of 180 secs.  Reconnects include jitter


rebar.config
```diff
@@ -30,7 +30,8 @@
         {jsonrpc2, {git, "https://github.com/novalabsxyz/jsonrpc2-erlang",
                  {branch, "master"}}},
         {xorf, {git, "https://github.com/helium/erlang-xorf.git",
-                      {branch, "main"}}}
+                      {branch, "main"}}},
+        {backoff, {git, "https://github.com/ferd/backoff.git", {branch, "master"}}}
```
Vagabond, line 34: backoff was already in the dep tree, why are you overriding it?

andymck: because it wasnt a dep of miner, but of another dep ?

Vagabond:
Shouldn't we increase the backoff if the MAX_BLOCK_AGE is too high, as well?

andymck:
> Shouldn't we increase the backoff if the MAX_BLOCK_AGE is too high, as well?

if the max block age is exceeded, the statem will reenter the setup state and thus we will hit the backoff

rebar.config
andymck: its not a prob but i think its good practise to include a dep if its required by the app

Vagabond: I guess I disagree, because it might break the behaviour of the library that originally pulled in the dep.

rebar.lock
```diff
@@ -3,7 +3,10 @@
   {git,"https://github.com/novalabsxyz/acceptor_pool",
        {ref,"56d676e00c11fd071a6bcc4059e3454960900af7"}},
   2},
- {<<"backoff">>,{pkg,<<"backoff">>,<<"1.1.6">>},2},
```
xandkar, line 6: Can just just depend on this same `1.1.6` package instead of master from git.

jeffgrunewald (approved):

rebar.config
andymck: for sake of haste i pulled the dep from miner

config/val.config.src
```diff
@@ -57,7 +57,9 @@
    %% we only really need the params below if this file is changed to specify a radio device
    %% as without one miner_lora is not started
    %% including the params anyway in case someone needs it in this env
-   {region_override, 'US915'}
+   {region_override, 'US915'},
+   %% max expodential backoff to apply to grpc reconnect
```
abhay, line 61:
```suggestion
   %% max exponential backoff to apply to grpc reconnect
```

config/val.config.src
```diff
@@ -82,6 +84,11 @@
                       header_table_size => 4096,
                       enable_push => 1,
                       max_concurrent_streams => unlimited,
```
abhay, line 86: do we want to change things as well?

andymck: we could but i doubt it is causing any probs...its the max num of streams per connection

config/val.config.src
```diff
@@ -82,6 +84,11 @@
...
+                      max_connections => 2500,                   %% the max connections permitted, after which connections will be refused
```
Vagabond, line 87: This seems low? I'm not sure what kind of numbers we normally see

config/val.config.src
```diff
@@ -57,7 +57,9 @@
    %% we only really need the params below if this file is changed to specify a radio device
    %% as without one miner_lora is not started
    %% including the params anyway in case someone needs it in this env
-   {region_override, 'US915'}
+   {region_override, 'US915'},
+   %% max exponential backoff to apply to grpc reconnect
+   {max_backoff, 128}
```
Vagabond, line 62: I think we discussed making this up to 10 minutes?

config/val.config.src
madninja: that does seem low since there's durable connections as well as the spew of connections as part of challenge handling

config/val.config.src
andymck: this is a multiplier to the default backoff of 5 secs, so 5 * 128 = 640 secs or just over 10 mins.  
Updated comment to explain better and also moved this config to sys config as its a clientside config not validator
ca29e9f79c837386f36ed16d1f32c9fcc37cc283

config/val.config.src
andymck: i have bumped to 4k connections here ca29e9f79c837386f36ed16d1f32c9fcc37cc283. 

Looking back on val ops channel, the spikes do seems to be around the 2.5k mark.

If you consider 1m hotspots spread across 1000 vals ( even tho we have much more than that ) it works out average of 1K connections per hotspot.

But maybe best to start high and work down from there

madninja (approved):

PR #1848, opened 2022-11-01
Use API for augmenting target block time calculation

mawdegroot:
This PR tries to mitigate the penalties that occur when a validator has a block history that leads to a different target block time than the majority of the other validators. `miner_blockhistory` fetches the block time and height from 50k blocks ago every 5 minutes and caches it for the target block time calculation to use. If the validator has sufficient block history itself it will not use `miner_blockhistory` and when `miner_blockhistory` provides erroneous values they will also be disregarded.

### DIFF
--- a/src/miner_blockhistory.erl
+++ b/src/miner_blockhistory.erl
@@ -33,7 +33,7 @@ handle_info(check, #state{url=URL}=State) ->
     case blockchain_worker:blockchain() of
         undefined ->
             lager:info("No chain, retry soon"),
-            {ok, schedule_check(State, 15)};
+            {ok, schedule_check(State, timer:seconds(15))};
         Chain ->
             Ledger = blockchain:ledger(Chain),
             case blockchain_ledger_v1:current_height(Ledger) of
--- a/src/miner_blockhistory.erl
+++ b/src/miner_blockhistory.erl
@@ -1,0 +1,97 @@
+-module(miner_blockhistory).
+
+-behavior(gen_server).
+
+-record(state, {
+    url,
+    historical_block=0,
+    historical_time=0,
+    last_retrieved_time=0
+}).
+
+-define(HISTORICAL_TIME_FETCH_INTERVAL, 300).
+
+-export([init/1, handle_info/2, handle_cast/2, handle_call/3]).
+
+-export([start_link/0, get_history/0]).
+
+start_link() ->
+    BaseURL = application:get_env(miner, api_base_url, "https://api.helium.io/v1"),
+    URL = string:concat(BaseURL, "/blocks/"),
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [URL], []).
+
+-spec get_history() -> {non_neg_integer(), non_neg_integer()} | undefined.
+get_history() ->
+    gen_server:call(?MODULE, get_history).
+
+%% gen_server callbacks
+
+init([URL]) ->
+    {ok, schedule_check(#state{url=URL}), 15}.
+
+handle_info(check, #state{url=URL}=State) ->
+    case blockchain_worker:blockchain() of
+        undefined ->
+            lager:info("No chain, retry soon"),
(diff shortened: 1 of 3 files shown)


src/miner_blockhistory.erl
```diff
@@ -0,0 +1,97 @@
...
+    gen_server:call(?MODULE, get_history).
+
+%% gen_server callbacks
+
+init([URL]) ->
+    {ok, schedule_check(#state{url=URL}), 15}.
+
+handle_info(check, #state{url=URL}=State) ->
+    case blockchain_worker:blockchain() of
+        undefined ->
```
mawdegroot, line 34: This could probably be handled better by correctly waiting for the chain to be available.

Vagabond: its not terrible, the other way is to subscribe to blockchain_events and wait for the `new_chain` event or whatever

src/miner_blockhistory.erl
```diff
@@ -0,0 +1,97 @@
...
+            lager:info("No chain, retry soon"),
+            {ok, schedule_check(State, 15)};
+        Chain ->
+            Ledger = blockchain:ledger(Chain),
+            case blockchain_ledger_v1:current_height(Ledger) of
+                {ok, CurrentHeight} ->
+                    StabilizationPeriod = application:get_env(miner, stabilization_period, 50000),
```
Vagabond, line 41: we default this to 0 in miner:set_next_block_timer, would be wise to be consistent here

src/miner_blockhistory.erl
```diff
@@ -0,0 +1,97 @@
...
+                    HistoryHeight = max(1, CurrentHeight - StabilizationPeriod),
+                    HeightURL = string:concat(URL, integer_to_list(HistoryHeight)),
+                    case httpc:request(get, {HeightURL, [{"user-agent", "ValidatorHistoricalHeight/0.0.1"}] }, [], [{body_format, binary}]) of
+                        {ok, {{_HttpVersion, 200, "OK"}, _Headers, Body}} ->
+                            try jsx:decode(Body, [{return_maps, true}]) of
+                                Json ->
+                                    case maps:get(<<"data">>, Json, undefined) of
```
Vagabond, line 48: You could probably do `maps:get(<<"time">>, maps:get(<<"data">>, Json, #{}), undefined)` and save yourself a case statement here

src/miner_blockhistory.erl
Vagabond: If you used the add_block event to schedule a check, instead of a 15 second loop (and maybe did a consensus membership check) you could get the exact historical block time you'd need for the next block?

src/miner_blockhistory.erl
```diff
@@ -0,0 +1,97 @@
...
+%% gen_server callbacks
+
+init([URL]) ->
+    {ok, schedule_check(#state{url=URL}), 15}.
+
+handle_info(check, #state{url=URL}=State) ->
+    case blockchain_worker:blockchain() of
+        undefined ->
+            lager:info("No chain, retry soon"),
+            {ok, schedule_check(State, 15)};
```
Vagabond, line 36: This is in milliseconds

src/miner_blockhistory.erl
```diff
@@ -0,0 +1,97 @@
...
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [URL], []).
+
+-spec get_history() -> {non_neg_integer(), non_neg_integer()} | undefined.
+get_history() ->
+    gen_server:call(?MODULE, get_history).
+
+%% gen_server callbacks
+
+init([URL]) ->
+    {ok, schedule_check(#state{url=URL}), 15}.
```
Vagabond, line 30:
```suggestion
    {ok, schedule_check(#state{url=URL}, 15)}.
```

Vagabond (approved):

PR #1852, opened 2022-11-03
add nonresponsive healthcheck to helium_gateway port

jeffgrunewald:
Add a periodic healthcheck to the port process managing the helium_gateway. starting from the init of the port, the managing process will schedule itself every interval (currently 10 minutes) to shell out to the helium_gateway cli with the command `os:cmd("helium_gateway -c config/dir info")`. if the rust gateway doesn't respond within 2 seconds or the response doesn't appear to be the expected response, the port managing process will kill itself, which will cause the supervisor to restart it. In the process of killing itself, the managing  process calls `kill -9` on the os pid of the helium_gateway so it should successfully restart as soon as the beam attempts to revive the port manager.


src/miner_gateway_port.erl
```diff
@@ -79,6 +82,25 @@ handle_info({'DOWN', Ref, port, _Pid, Reason}, #state{port = Port, monitor = Ref
                    State#state.tcp_port),
     ok = miner_gateway_ecc_worker:reconnect(),
     {noreply, NewState};
+handle_info(gw_healthcheck, State) ->
+    Me = self(),
+    Healthcheck = fun() -> Output = os:cmd(gateway_bin() ++ " -c " ++ gateway_config_dir() ++ " info"), Me ! Output end,
```
xandkar, line 87: I'd use the following alternative to `os:cmd/1`, so that we can check the exit code: https://gist.github.com/xandkar/20ef60a5ec2faaa069d71ee0e073f88e

Makefile
```diff
@@ -11,8 +11,8 @@ endif
 
 GRPC_SERVICES_DIR=src/grpc/autogen
 
-GATEWAY_RS_VSN ?= "v1.0.0-alpha.31"
-GWMP_MUX_VSN ?= "v0.9.4"
+GATEWAY_RS_VSN ?= "8bf086618787b7eacf64abefdadf65a383e4c954"
```
vihu, line 14: We're switching to commit hash instead of tag?

src/miner_gateway_port.erl
```diff
@@ -79,6 +82,25 @@ handle_info({'DOWN', Ref, port, _Pid, Reason}, #state{port = Port, monitor = Ref
...
+    Checker = spawn(Healthcheck),
+    receive
+        Response ->
+            case string:str(Response, "gateway") of
```
xandkar, line 91: Would be nice if we could pattern match this instead. Maybe something like:
```erl
case iolist_to_binary(Response) of
    <<"gateway", _/binary>> -> ...;
    <<_/binary>> -> ...
end
```

and even better with `case {ExitCode, iolist_to_binary(Response)} of`

Makefile
jeffgrunewald: we've been using commit hash whenever there was desire to avoid tagging; @madninja doesn't want to generate a new tag and kick off a redeploy/update of all light gateways just for a miner release

src/miner_gateway_port.erl
jeffgrunewald: we don't really care about the contents of the output so much as that we _get_ the output; the condition we're trying to guard against is the helium_gateway getting hung such that calling the above command never returns _anything_

xandkar: oh, sweet - that's even simpler then - just check exit code.

src/miner_gateway_port.erl
```diff
@@ -79,6 +82,15 @@ handle_info({'DOWN', Ref, port, _Pid, Reason}, #state{port = Port, monitor = Ref
                    State#state.tcp_port),
     ok = miner_gateway_ecc_worker:reconnect(),
     {noreply, NewState};
+handle_info(gw_healthcheck, State) ->
+    case healthcheck_cmd() of
+        {0, _Output} ->
+            erlang:send_after(?HEALTHCHECK_INTERVAL, self(), gw_healthcheck),
+            {noreply, State};
+        {Exit, Reason} when Exit /= 0 ->
```
xandkar, line 90:
```suggestion
        {Exit, Reason} ->
```
0 check is already covered above, so this is redundant.

src/miner_gateway_port.erl
```diff
@@ -131,6 +143,28 @@ cleanup_port(#state{port = Port} = State) ->
     os:cmd(io_lib:format("kill -9 ~p", [State#state.os_pid])),
     ok.
 
+healthcheck_cmd() ->
```
xandkar, line 146:
```suggestion
-spec healthcheck_cmd() -> {non_neg_integer(), iolist()}.
healthcheck_cmd() ->
```

src/miner_gateway_port.erl
```diff
@@ -131,6 +143,28 @@ cleanup_port(#state{port = Port} = State) ->
...
+        {PortId, eof} ->
+            erlang:port_close(PortId),
+            receive
+                {PortId, {exit_status, ExitCode}} ->
+                    {ExitCode, lists:reverse(Data)}
+            end
+    after
+        ?HEALTHCHECK_TIMEOUT ->
+            erlang:port_close(PortId),
+            {1, timeout}
```
xandkar, line 165:
```suggestion
            timeout
```

xandkar (approved):

PR #1855, opened 2022-11-07
Fix gwrs-build and add new gwrs settings

Vagabond:


priv/gateway_rs/settings.toml
```diff
@@ -1,2 +1,8 @@
 ## api listening port override
 api = 4468
+
+[poc]
+entropy_uri = "https://entropy.helium.io/entropy"
+ingest_uri = "http://mainnet-pociot.helium.io:9080"
+interval = 600
+ingest_token = "dev"
```
jeffgrunewald, line 8:
```suggestion
interval = 600
```
the entropy_uri and ingest_uri are just replicating the defaults so those can be omitted and the lora iot network doesn't use an ingest_token./

the interval can be set as above or if you prefer you could set it in the miner_gateway_port as the environment variable `GW_POC_INTERVAL=600`; either way.

priv/gateway_rs/settings.toml
```diff
@@ -1,2 +1,8 @@
 ## api listening port override
 api = 4468
+
+[poc]
+entropy_uri = "https://entropy.helium.io/entropy"
+ingest_uri = "http://mainnet-pociot.helium.io:9080"
+interval = 600
```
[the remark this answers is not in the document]
Vagabond: seconds, this is just for testing we can increase it if things look good

priv/gateway_rs/settings.toml
Vagabond: Removed the other values

jeffgrunewald (approved):

PR #1865, opened 2022-12-08
Stage a possible patch to PoC spreading factor choice.

ke6jjj:


Vagabond:
shouldn't this be done via the region parameters chain var? @vihu ?

vihu:
> shouldn't this be done via the region parameters chain var? @vihu ?

Yeah, that would be ideal. @ke6jjj @mikev can you take a look at https://github.com/helium/blockchain-core/blob/master/test/blockchain_region_test.hrl file and update it according to the new spec, I assume only the spreading values need to change there? We can then use the changes and update the `region_as923_1_params`, `region_as923_2_params` etc accordingly via chain var.

ke6jjj:
We are structuring the change this way precisely to _avoid_ doing a chain var. The fact of the matter is that chain var changes are very risky at the moment and I don't feel we're comfortable making them so close to a holiday period.

vihu:
> We are structuring the change this way precisely to _avoid_ doing a chain var. The fact of the matter is that chain var changes are very risky at the moment and I don't feel we're comfortable making them so close to a holiday period.

Fair point. However, this would be a chain var for `region_X_params` and not a region change. I believe those are not soo risky since there's no cache cleanup required. Thoughts @Vagabond ?

PR #1887, opened 2023-01-23
If the `isolate_var_txns` var is set, put var txns in their own block

Vagabond:
Depends on helium/blockchain-core#1519

### DIFF
--- a/src/miner.erl
+++ b/src/miner.erl
@@ -716,7 +716,14 @@ select_transactions(Chain, Txns, {ElectionEpoch0, EpochStart0}, BlockHeightCurr,
                      ),
                 {Epoch, BlockHeightNext, Transactions, InvalidTransactions};
             _ ->
-                {ElectionEpoch0, EpochStart0, ValidTransactions, InvalidTransactions}
+            %% try to isolate the var txn in its own block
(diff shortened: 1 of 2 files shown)


src/miner.erl
```diff
@@ -716,7 +716,14 @@ select_transactions(Chain, Txns, {ElectionEpoch0, EpochStart0}, BlockHeightCurr,
                      ),
                 {Epoch, BlockHeightNext, Transactions, InvalidTransactions};
             _ ->
-                {ElectionEpoch0, EpochStart0, ValidTransactions, InvalidTransactions}
+            %% try to isolate the var txn in its own block
+            case {blockchain:config(?isolate_var_txns, Ledger), lists:filter(fun(T) -> blockchain_txn:type(T) == blockchain_txn_vars_v1 end, ValidTransactions)} of
+                {{ok, true}, [VarTxn|_]} ->
+                    %% take the first var txn found as the only txn in a block
+                    {ElectionEpoch0, EpochStart0, [VarTxn], InvalidTransactions};
+                _ ->
```
vihu, line 724: It would fall back here if the var is set but there are no var txns correct?

vihu (approved):

src/miner.erl
Vagabond: yes, if *either* condition is false it falls back.

